PR #6, opened 2026-06-12
topbar: single Sign in button

strelov1:
Drop the separate Register button in the top bar; the auth dialog's own footer toggle switches to register, so one entry point suffices. Unrelated to PR #4 — split out as its own change.


strelov1:
Superseded: the single Sign-in button (dropping the separate Register button) already landed on main via PR #5 (add-oauth-sign-in), which collapsed the two buttons into one on the same lines and additionally introduced the UserMenu and OAuth error handling. Current main already shows a single 'Sign in' button, so this change is a no-op and conflicts only because it touches the same lines. Closing as already-done.

PR #321, opened 2026-07-01
refactor(web): move header-search visibility to TopBar

strelov1:
## Summary

Follow-up cleanup to #319. The "hide header search on /jobs & /companies" logic lived inside `HeaderSearch` as an `{#if !hidden}` wrap, which left the inner markup mis-indented and kept the component (with its window keydown/click listeners) mounted-but-hidden on those pages.

Hoist the route decision to `TopBar` (the layout owner): it renders `<HeaderSearch/>` or an empty `flex-1` spacer. `HeaderSearch` drops the route coupling (`page` import + `hidden` derived + wrap) and its markup returns to natural indentation. Behavior is identical (search hidden on the two list pages, shown everywhere else); on those pages the component now simply isn't mounted (no dead listeners).

## Test Plan

- [x] `svelte-check` 0/0, eslint clean (fresh origin/main base)

### DIFF
--- a/web/src/lib/components/HeaderSearch.svelte
+++ b/web/src/lib/components/HeaderSearch.svelte
@@ -1,6 +1,5 @@
 <script lang="ts">
   import { goto, afterNavigate } from '$app/navigation';
-  import { page } from '$app/state';
   import { resolve } from '$app/paths';
   import { Search, X } from '@lucide/svelte';
   import { api } from '$lib/api';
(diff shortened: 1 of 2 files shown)


strelov1:
Superseded: reworking header search into a single context-aware field (keep it on /jobs & /companies, drive their list; drop the page q inputs) — the hide/visibility approach is being replaced.

PR #932, opened 2026-07-19
fix(resolve-url): route proxied providers through the egress proxy

strelov1:
## Problem

Two habr_career jobs (and others) show empty descriptions on prod even though the origin pages have them, e.g. [1000167118](https://career.habr.com/vacancies/1000167118).

Root cause: `cmd/resolve-url` builds its linksource registry with a **direct-only** client (`sources.NewClient()`) and never applies the proxy egress. So resolving a single URL for a **proxied provider** (`habr_career`, `geekjob`) goes out on the prod datacenter IP. Those detail pages sit behind **Qrator**, which challenges that IP — the `JobPosting` ld+json parse fails and the description comes back empty. The board crawl (`cmd/ingest`) doesn't hit this: it routes these providers through `SOURCES_PROXY_URL` via `sources.ApplyProxyEgress`.

Verified on host-2: bare/browser UA curl to the habr detail page returns 200 + JobPosting (direct **and** via proxy), while `resolve-url`'s Go client is challenged → `no JobPosting ld+json`. The discriminator is the datacenter-IP + Go-`net/http` TLS fingerprint that Qrator scores; the residential proxy clears it — exactly what `ApplyProxyEgress` already exists for.

## Fix

Mirror the crawl's selective proxy egress on the single-URL path:

- `sources.IsProxied(provider)` — exported predicate over the proxied allowlist.
- `sources.ProxyClientFromEnv()` — builds the proxy client from `SOURCES_PROXY_URL` (nil when unset, error when set-but-invalid); `ApplyProxyEgress` now reuses it.
- `linksource.AllWithProxyEgress(direct, proxy)` — routes only proxied-allowlist adapters through the proxy, keeping every other adapter (and the generic weblink fallback) on the direct client. A nil proxy leaves everything direct.
- `cmd/resolve-url` wires `SOURCES_PROXY_URL` through, guarding the typed-nil-interface trap.

## Tests

- `linksource`: proxied providers get the proxy client, the rest stay direct; nil proxy stays fully direct.
- `sources`: `ProxyClientFromEnv` nil-when-unset / builds-when-set / errors-without-leaking-creds.
- Full suite green (`go test ./...`).

Selective by design — if the residential proxy is down/quota-exhausted (a recurring incident), direct providers like greenhouse/ashby keep working.

### DIFF
--- a/cmd/resolve-url/main.go
+++ b/cmd/resolve-url/main.go
@@ -63,11 +63,28 @@ func run() int {
 		idx = search.NewClient(cfg.MeiliURL, cfg.MeiliKey)
 	}
 
+	// Route the proxied link-source adapters (habr_career/geekjob et al.) through the egress
+	// proxy when one is configured (SOURCES_PROXY_URL), exactly as cmd/ingest does for the
+	// crawl — their detail pages sit behind a WAF (Qrator) that challenges the prod datacenter
+	// IP, so a direct single-URL resolve gets an empty description. A set-but-invalid value
+	// fails the run here rather than silently resolving on the blocked direct IP.
+	proxy, err := sources.ProxyClientFromEnv()
+	if err != nil {
+		log.Printf("resolve-url: %v", err)
+		return 1
+	}
+	direct := sources.NewClient()
+	// A nil *sources.Client wrapped in the interface would read as non-nil; keep the
+	// interface value nil so AllWithProxyEgress leaves every adapter on the direct client.
+	var proxyClient linksource.Client
+	if proxy != nil {
+		proxyClient = proxy
+	}
 	// The generic resolver is appended AFTER the host-scoped adapters (so a known ATS is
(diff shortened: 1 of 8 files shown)


strelov1:
Closing: superseded by the real diagnosis. The habr empty-description issue is a **parser bug** (schema.org `jobLocation` object-vs-array dropped the whole JobPosting incl. description), fixed in #938 — not the egress IP.

Also, measured from the prod datacenter IP: Qrator challenges **every Go TLS stack** (native `net/http` and utls-Chrome, direct **and** via this proxy) 0/6, while curl/OpenSSL passes 100%. So routing resolve-url through the proxy has **no payoff** — habr/geekjob are TLS/JA4-fingerprint-gated, not IP-gated, and no pure-IP-blocked provider has a linksource adapter. The proxy-egress mechanism was correct but moves nothing today.

PR #1071, opened 2026-07-22
feat(web): overlapping country-flag stack + Discussion link in job/company header

strelov1:
## What

- **CountryFlagStack** — avatar-style cluster of round country flags that overlap (leftmost on top, ring-card separators, `+N` overflow), so a many-country remote role stays a compact row instead of wrapping.
  - JobRow: eligible-country flags in the signal row (display-only — the card is a link).
  - JobView sidebar: replaces the wrapping country flag row with the stack (linked to per-country filters).
- **Discussion link** lifted into the job/company header top row (right-aligned beside the company name; under Subscribe on the company header); removed the duplicate that rendered in the header body.

## Notes
3 web-only commits. No API/DB changes.

### DIFF
--- a/web/src/lib/components/CompanyHeader.svelte
+++ b/web/src/lib/components/CompanyHeader.svelte
@@ -48,20 +48,19 @@
       {/if}
     </div>
     <!-- flex-1 above lets the name/tagline use the full width; the follow CTA stays
-         top-right and never shrinks (it collapses to an icon on mobile itself). -->
-    <div class="shrink-0">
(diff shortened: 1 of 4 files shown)


strelov1:
Closing: this branch is stale — 19 commits behind main, and all its work (CountryFlagStack, the Discussion-link header move) is already merged into main via other PRs. The only net diff vs current main is a 3-line JobView header that would *revert* newer layout work. Nothing to ship here.

PR #1092, opened 2026-07-23
DRAFT: DS Phase 3: Design tokens

gearsandcode:
Phase 3 of the design-system build ([gearsandcode/freehire-design#5](https://github.com/gearsandcode/freehire-design/issues/5)).

DTCG-format token JSON + Style Dictionary build generating CSS custom properties, replacing hand-written values in `app.css` as the source of truth.

## What this produces

- **8 DTCG token files** in `design-system/tokens/`: color (light), color-dark, spacing, typography, radius, shadow, motion, z-index. Every token has ``, ``, and ``.
- **Style Dictionary build** (`design-system/scripts/build-tokens.mjs`): custom `shadow/css` transform, two SD instances (light → `:root`, dark → `.dark`), alias `{brand-ring}` resolves per-theme.
- **Generated CSS** in `design-system/dist/tokens-{light,dark}.css` (committed — `file:` dependency has no publish step).
- **`app.css` rewired**: hand-authored `:root` + `.dark` blocks replaced with `@import` from `freehire-design-system/dist/tokens-{light,dark}.css`.

## Token coverage

- **Color:** 23 semantic tokens ported from `app.css` (bg/fg/card/popover/primary/secondary/muted/accent/destructive/border/input/ring + brand family). Dark overrides for all 23.
- **Spacing:** 13-step scale (0-24), Tailwind-aligned.
- **Typography:** 2 font families + 8 font sizes + 4 font weights + 3 line heights.
- **Radius:** base + sm/md/lg/xl derivations.
- **Shadow:** 4-step elevation scale (sm/md/lg/xl).
- **Motion:** 3 durations + 3 easing curves.
- **Z-index:** 6-step stacking scale.

## Verified locally
- `pnpm tokens:build` regenerates CSS matching existing `app.css` values + new categories
- `pnpm run check` (0 errors, 11 pre-existing warnings)
- `pnpm run build` (green)
- Every color token has both light and dark values

**Depends on PRs #1087 (phase 1) + #1088 (phase 2) being merged first.**

Refs: gearsandcode/freehire-design#5

PR #1088, opened 2026-07-23
DS Phase 2: Package boundary and tooling migration

gearsandcode:
Phase 2 of the design-system build ([gearsandcode/freehire-design#4](https://github.com/gearsandcode/freehire-design/issues/4)).

Pure repo plumbing — stands up `design-system/` as an independent pnpm package and migrates `web/` from npm to pnpm. **Depends on phase 1 (#1087) being merged first** — the `design-system/docs/` directory it creates doesn't exist on main without it.

## Changes

- **design-system/ scaffold:** `package.json` with placeholder scripts, `pnpm-lock.yaml`, empty `src/` + `tokens/` dirs.
- **web/ → pnpm:** deleted `package-lock.json`, added `pnpm-lock.yaml` + `pnpm-workspace.yaml`, added `freehire-design-system: file:../design-system` dependency.
- **CI:** two pnpm jobs — `design-system` and `web` (`needs: [design-system]`).
- **Dockerfile:** pnpm via corepack, build context → repo root, `ds-build` stage.
- **docker-compose.yml:** `web` build context `./web` → `.`.
- **CONTRIBUTING.md / AGENTS.md / PR template / API docs script:** all npm refs → pnpm.

## Verified locally
- `pnpm install --frozen-lockfile` in both packages
- `pnpm run check` (0 errors, 11 pre-existing warnings)
- `pnpm run build` (green)

PR #1087, opened 2026-07-23
DS Phase 1: Investigate and inventory

gearsandcode:
Phase 1 of the design-system build ([gearsandcode/freehire-design#3](https://github.com/gearsandcode/freehire-design/issues/3)).

Read-only investigation checked into `design-system/docs/`. No code changes.

## Deliverables

- **`design-system/docs/component-inventory.md`** — all 99 `.svelte` components under `web/src/lib/components/` classified systemized (23) vs bespoke (76), with one-line reasons. Seven migration-candidate patterns identified.
- **`design-system/docs/icon-inventory.md`** — every icon in `web/`: 172 `@lucide/svelte` usages across 48 files + 13 bespoke SVG files.
- **`design-system/docs/inventory-summary.md`** — stack summary, existing 4 primitives, token gaps vs `app.css`, infrastructure touch points for phase 02, and the low-churn decision to keep `/ui` as a thin re-export.

## Exit check
- [x] Components inventoried and classified
- [x] Primitives-in-disguise flagged
- [x] `app.css` tokens noted + gaps identified
- [x] Lint/TS config checked
- [x] Svelte 5 runes confirmed as house style
- [x] Icon audit complete
- [x] No existing token-build tooling (confirmed)
- [x] CI/Docker/compose/CONTRIBUTING read in full

PR #1096, opened 2026-07-23
fix(ci): add root package.json for pnpm/action-setup version detection

gearsandcode:
Permanent fix for the recurring `No pnpm version is specified` CI failure.

## Root cause

`pnpm/action-setup@v4` runs from the **repo root** — `working-directory` in `defaults.run` only affects `run:` steps, not `uses:` steps. Without a root `package.json`, the action can't find the `packageManager` field and fails. The `version: 11` workaround in the workflow config only exists on branches that have it — every new branch from `main` hits the same failure.

## Fix

Add a root-level `package.json` with `packageManager: "pnpm@11.7.0"`. The action always finds it at the repo root, regardless of working-directory. Every branch from `main` inherits this — no more cherry-picking the fix onto every PR.

**Merge this first** — it unblocks all the DS phase PRs (#1087, #1088, #1092) and prevents the failure on future branches.


gearsandcode:
Closing — folding this into PR #1087 (ds/phase-1-investigate) instead of a standalone PR.

PR #1137, opened 2026-07-26
feat(assistant): BYOK — setup instructions and a where-it-runs badge

strelov1:
Frontend for the assistant running on the user's own machine. Server side is already deployed (freehire-agent #17, #19–#23).

## What changes

**A badge in the chat header** showing where the assistant runs: *Your machine* when a runner is connected, *Our servers* otherwise, and a warning when a runner is required but missing. Polled every 10s — the runner connects to the agent backend, not to this page, so there is nothing to subscribe to. A failed poll keeps the last known state rather than flickering to disconnected.

**Setup instructions instead of an error.** When the assistant may only run on the user's machine and none is connected, the backend answers `409 no_device`. That is an unfinished setup, not a failure, so the UI shows the five commands to run. Wired into both surfaces that open a session: the assistant chat and the tailoring workspace.

## One detail worth the extra type

`NoDeviceError` is its own class rather than a message check. Otherwise the two cases would be told apart by string matching at the call site, and sooner or later someone would show installation instructions in response to a network blip.

## Checks

`svelte-check` 0 errors, eslint clean, 384 tests pass, production build succeeds.


strelov1:
Superseded by a clean branch: this one was cut from an unrelated feature branch and picked up its changes to `design-system/`. Reopened as a branch off `main` with only the five assistant files.

PR #1171, opened 2026-07-28
perf(ci): pr-smoke builds only the server binary it boots

strelov1:
The lever left over from #1170. The app image's `go build` was **66.5 s of pr-smoke's 131 s** stack build, and twelve of the thirteen binaries it produces are run-once workers the smoke never invokes.

## How

`WORKER_PKGS` is a build arg whose **default is the full list**, so an ordinary `docker build` — prod deploys, `make up` — still produces the complete image. The one override lives in `perf/docker-compose.smoke.yml` and is applied through `COMPOSE_FILE`, so:

- every `docker compose` call in the job picks it up without touching a single command in the workflow, and
- compose keeps building `app` and `web` **concurrently**. A separate build step would serialize them and hand back more than this saves — that's what went wrong in #1170's first attempt.

## The trade, stated plainly

The smoke no longer boots a byte-identical prod image. Same base image, same runtime stage, same server binary, same `migrations/` and `typst` — it differs only by binaries that image would never run during the smoke. "Does every cmd package still compile" is already covered by `go vet ./...` in the backend job, and releases build the real image.

## Verified locally

| build | `/app` contents |
|---|---|
| `docker build .` (prod path) | all 13 binaries + `migrations/` + `typst` |
| with the overlay | `hire`, `migrations/`, `typst` — and the server starts (fails fast on the missing JWT_SECRET) |

Note the fail-safe direction: if compose ever stopped passing the empty arg through, the Dockerfile default takes over and the smoke just builds everything, as it does today.

**The CI number is what decides this.** Expected ~25–30 s off the step, not the full ~40 s — `web`'s image path (~90–105 s: two `pnpm install`s + a 53 s `pnpm run build`) becomes the constraint. If the run says otherwise, this goes the way of #1170's cache experiment.

### DIFF
--- a/.github/workflows/perf.yml
+++ b/.github/workflows/perf.yml
@@ -34,6 +34,10 @@ jobs:
   pr-smoke:
     if: github.event_name == 'pull_request'
     runs-on: ubuntu-latest
+    env:
+      # Every `docker compose` below picks up the overlay, which builds the app
+      # image without the twelve worker binaries this smoke never invokes.
+      COMPOSE_FILE: docker-compose.yml:perf/docker-compose.smoke.yml
     steps:
       - uses: actions/checkout@v7
         with:
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,13 +14,18 @@ COPY . .
 # Two invocations rather than one per binary: with `-o /out/` (a directory) Go
 # names each binary after its cmd dir — which is already how they're named — and
 # building them together shares one dependency compile and links them in
-# parallel. Cold-cache: 26s -> 17s locally, 72s -> ~50s on a CI runner. Only the
+# parallel. Cold-cache: 26s -> 17s locally, 72s -> 66s on a CI runner. Only the
 # server needs an output name of its own.
+#
+# WORKER_PKGS is the full list by default, so an ordinary `docker build` still
+# produces the complete image. The ONLY thing that overrides it is the CI page
+# smoke (perf/docker-compose.smoke.yml), which boots nothing but the server and
+# has no use for the workers' compile time.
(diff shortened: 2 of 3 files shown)


strelov1:
Closing — measured, and it isn't worth it.

Run [30329850111](https://github.com/strelov1/freehire/actions/runs/30329850111): the app image's `go build` went **66.5 s → 64.1 s**, and the whole stack build 131 s → 128 s, which is inside run-to-run noise. **The twelve worker binaries cost ~2.5 s.** Everything else in that 64 s is compiling the dependency graph and linking the server — exactly what the smoke needs by definition.

My local numbers (15–25 s server-only vs 29–31 s for all thirteen) were thermal noise on a laptop; back-to-back repeats disagreed with each other by 9 s, which should have been the tell before I built anything.

So the trade is real (the smoke stops booting a byte-identical prod image, plus a CI-only build knob to maintain) and the return is ~2.5 s. Not merging.

Where that leaves pr-smoke's 3m33s, all of it irreducible without cutting what the gate actually checks:

| | |
|---|---|
| `go build` of the server (app image) | 64 s |
| `pnpm run build` of the SSR frontend | 50 s |
| apt (poppler + ca-certificates) | 18 s |
| two `pnpm install`s | 26 s |
| k6 itself | 65 s |

app (~100 s) and web (~95 s) build concurrently, so both would have to shrink together to move the number. A layer cache was measured negative in #1170 (buildx's export/import into the daemon costs ~65 s, more than the ~30 s of cacheable deps layers). The remaining option is building the Go binary on the runner with setup-go's warm cache and injecting it into the image — ~28 s off, for a CI-only build path and a real fidelity loss. Not obviously worth it either.

PR #1172, opened 2026-07-28
fix(cv): restore the narrow cv-scope gate the tailoring agent authenticates with

strelov1:
## What broke

The handler split (#1158) moved the CV routes from `cvKeyAuth` onto `mw.key`.
`mw.key` is `RequireAuthOrKey`, which is `RequireAuthOrScopedKey` with **no**
admitted scopes — so only a full-scope key passes, and anything narrower gets
`403 insufficient scope`.

The tailoring bootstrap mints its agent a key at the narrow `cv` scope
(`mintTailoringKey`, `cv_tailor.go`). Every request that agent makes has
therefore been rejected since the split:

| Route | Driven by |
|---|---|
| `GET /me/cvs/:id` | `freehire cv get` |
| `GET /me/cvs/:id/pdf` | `freehire cv render` |
| `PATCH /me/cvs/:id` | `freehire cv edit` |
| `PUT /me/cvs/:id/session` | session bookkeeping |
| `GET /me/cvs/:id/tailor-context` | `freehire cv context` |

CV tailoring over the CLI is broken in production. Authoring in the browser is
unaffected — those routes are cookie-gated and were not touched.

## Why the tests missed it

`cv_tailor_integration_test.go` builds its own app and mounts the routes on
`RequireAuthOrScopedKey(..., ScopeCV)`. Its comment states the intent plainly —
"a harness on the full-scope-only middleware would reject the very credential
the flow issues" — but because it never exercises the real router, it asserts
the intent rather than the wiring, and stayed green while production diverged.

## The fix

Five routes back onto `mw.cvKey`, plus a test that drives `register()` itself
and pins the gate behind each CV route, so a future move shows up as a failure
instead of a 403 in production.

Test plan: `go test ./...` green; the new `TestCVRegister_TailoringRoutesAdmitTheNarrowCVKey`
fails on `main` with exactly those five routes and passes here.

### DIFF
--- a/internal/handler/cv.go
+++ b/internal/handler/cv.go
@@ -61,21 +61,23 @@ func (h *cvHandlers) register(api fiber.Router, mw middleware) {
 	api.Get("/cv-templates", mw.cookie, h.ListCVTemplates)
 	api.Get("/me/cvs", mw.cookie, h.ListCVs)
 	api.Post("/me/cvs", mw.cookie, h.CreateCV)
-	// Read + render accept a key too (keyAuth), so the tailoring agent's CLI can fetch a CV
-	// and its PDF; mutations stay cookie-only (POST/PUT/DELETE — the browser owns authoring).
-	api.Get("/me/cvs/:id", mw.key, h.GetCV)
+	// Read + render accept a key too, so the tailoring agent's CLI can fetch a CV and its
+	// PDF; mutations stay cookie-only (POST/PUT/DELETE — the browser owns authoring).
+	// cvKey, not key: the bootstrap mints the agent a narrow `cv`-scoped key, and key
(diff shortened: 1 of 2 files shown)


strelov1:
Closing: the premise does not hold on current `main`.

I raised this from a worktree branched off a stale local `main`, five commits behind. On the real `main`, #1165 moved the assistant in-process and `mintTailoringKey` went with it — `git grep mintTailoringKey origin/main` returns nothing. No credential is minted at the `cv` scope any more, so no caller can hit the 403 this fixes. `cvKeyAuth` is still constructed (`handler.go:338`) but is now vestigial.

The route test was the useful half — it pins the gate per route against the real `register()`, which the existing integration test cannot do because it mounts its own app. If the `cv` scope ever gets a live credential again, that test is worth reviving.

PR #1177, opened 2026-07-28
fix(inbox): the Gmail callback is a browser navigation, not an XHR

strelov1:
## The report

```
GET https://freehire.dev/api/v1/me/gmail/callback?state=…&code=…
→ {"error": "not authenticated"}
```

A user pressed **Connect Gmail**, consented at Google, and landed on a JSON error page with no way back into the app.

## Root cause

Two separate faults, one of them not in this repo.

**1. The dead end (fixed here).** `/me/gmail/callback` was mounted on `mw.cookie` (`RequireAuth`), which answers 401 *before* the handler runs. `GmailCallback` has always handled the unauthenticated case itself — `redirect("gmail_error=auth", …)` — but that branch has been unreachable since the feature landed in #618. An OAuth callback is a top-level browser navigation; it must never render JSON.

**2. The trigger (ops, NOT in this PR).** Prod builds the Gmail `redirect_uri` from a fixed `FRONTEND_ORIGIN` (`internal/gmailsync/connector.go:42`), which still points at the legacy `freehire.dev` origin. Sessions are scoped `.freehire.me` (`internal/auth/cookie.go:59`) and the state cookie is host-only — so neither cookie survives the hop to `.dev`, and the callback arrives anonymous.

Evidence gathered against prod:

| Probe | Result |
|---|---|
| `GET https://freehire.dev/` | `301 → https://freehire.me/` — the SPA is not browsable on `.dev` |
| `GET https://freehire.dev/api/v1/me/gmail` | `401` — `/api/` on `.dev` is **not** redirected, it reaches the app |
| `/api/v1/auth/oauth/google/start` on both hosts | `redirect_uri` is per-host — sign-in already solved this (`requestOrigin`, `internal/handler/oauth.go:25`) |

The codebase documents this exact failure class at `internal/handler/oauth.go:35-43` — "the state cookie is set on the host the flow started from, while the callback is sent to the canonical origin, so the state can never match." Sign-in was taught the lesson; the Gmail connector never was.

## What changed

- **`auth.OptionalCookieAuth`** — attaches a cookie session when present, passes through anonymously otherwise. Cookie-only on purpose: `handler.go:343` declares the inbox a surface a leaked API key must not reach, and a browser returning from a provider redirect carries a cookie or nothing, so admitting Bearer would widen a key's reach for no gain.
- **The callback route** moves to it, which brings the handler's own `gmail_error=auth` redirect back to life.
- **`InboxView.svelte`** finally reads the outcome markers. `?gmail=connected` and `?gmail_error=` have been arriving at `/my/inbox` since #618 and going nowhere — even a *successful* connect returned the user to a silent page. Now both surface as a banner (not the fatal `error` screen — the inbox itself is fine), then the URL is cleaned, mirroring `TopBar`'s `?auth_error` handling.
- **Two tests** mounting the real `register()` route table, so the middleware the callback is actually served behind is what is exercised. Both failed before the fix.

## Still required to actually close the bug

This PR fixes the dead end, not the trigger. Ordered, because the first step is a prerequisite:

1. **Google Cloud Console** — confirm `https://freehire.me/api/v1/me/gmail/callback` is an Authorized redirect URI. If only `.dev` is registered, changing env below fails earlier with `redirect_uri_mismatch`.
2. **Prod env** — `FRONTEND_ORIGIN=https://freehire.me`, `SERVED_HOSTS=freehire.me`, `COOKIE_DOMAIN=.freehire.me`.
3. **nginx on the host** — fold `/api/` on `.dev` into the 301 the rest of that domain already gets.

Narrowing `COOKIE_DOMAIN` signs out anyone still holding a `.freehire.dev` cookie (and `ClearTokenCookie` cannot purge it — the domain no longer matches), and breaks any client with `.dev` hardcoded.

## Verification

`go build ./...`, `go vet ./...`, `gofmt -l` clean · `go test ./...` no failures · `pnpm lint` exit 0 (the two InboxView warnings are pre-existing, lines 391/394) · `pnpm build` exit 0.

### DIFF
--- a/internal/auth/middleware.go
+++ b/internal/auth/middleware.go
@@ -73,6 +73,27 @@ func RequireAuth(iss *Issuer, versions TokenVersionLoader) fiber.Handler {
 	}
 }
 
+// OptionalCookieAuth returns middleware that attaches the caller's user id when a
+// valid session cookie is present and otherwise passes through anonymously. It is
+// OptionalAuth without the Bearer path: for a surface a leaked API key must not
+// reach (the mail inbox), admitting a key here would widen that key's reach for no
+// gain — the only caller is a browser returning from a provider redirect, which
+// carries a cookie or nothing.
+//
+// The OAuth-style callbacks mount this rather than RequireAuth: they are top-level
+// navigations, so a 401 renders JSON into the browser and strands the user, while an
+// anonymous pass-through lets the handler redirect back with an error marker.
+func OptionalCookieAuth(iss *Issuer, versions TokenVersionLoader) fiber.Handler {
+	return func(c *fiber.Ctx) error {
+		if token := c.Cookies(CookieName); token != "" {
+			if id, ok := resolveSession(c, iss, versions, token); ok {
+				c.Locals(LocalsUserID, id)
+			}
+		}
+		return c.Next()
+	}
+}
+
 // resolveSession validates a cookie token and confirms it was not revoked, returning the
 // user id it authenticates. A version-load failure (deleted account, database trouble)
 // fails closed, matching RequireRole: an unverifiable session is not a session.
--- a/internal/handler/gmail.go
+++ b/internal/handler/gmail.go
@@ -62,7 +62,12 @@ func (h *inboxHandlers) register(api fiber.Router, mw middleware) {
 	api.Post("/me/emails/:id/reject", mw.cookie, h.RejectEmailLink)
 	if h.gmailReady() {
 		api.Get("/me/gmail/connect", mw.cookie, h.GmailConnect)
-		api.Get("/me/gmail/callback", mw.cookie, h.GmailCallback)
+		// The callback is the browser returning from Google, not an XHR — so it is
+		// mounted on optionalCookie, not cookie. Under RequireAuth a session that did
+		// not survive the round-trip (expired mid-consent, or a callback landing on a
+		// host the cookie is not scoped to) renders a JSON 401 into the address bar and
+		// strands the user; GmailCallback answers that case itself with a redirect.
+		api.Get("/me/gmail/callback", mw.optionalCookie, h.GmailCallback)
 		api.Post("/me/gmail/sync", mw.cookie, h.SyncGmail)
 	}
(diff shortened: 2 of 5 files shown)


strelov1:
**Shipped and verified on prod.** The ops trigger turned out to be more specific than this PR's description guessed, so recording it here.

`FRONTEND_ORIGIN` *was* already flipped to `https://freehire.me` — but only in `/opt/freehire/.env`. The `freehire-api@<color>` unit loads four EnvironmentFiles and the later ones win:

```
/opt/freehire/.env               FRONTEND_ORIGIN=https://freehire.me   <- flipped 2026-07-25
/opt/freehire/env/hire-api.env   FRONTEND_ORIGIN=https://freehire.dev  <- overrode it, missed
live process (/proc/PID/environ) FRONTEND_ORIGIN=https://freehire.dev  <- what the connector used
```

So the 2026-07-25 flip never reached the API. Everything else reading `FrontendOrigin` (mail footers, notify links, referral cabinet) had been serving `.dev` for three days unnoticed, hidden by the `.dev` -> `.me` 301 on HTML paths. Gmail was the one path where `/api/` is deliberately not redirected, so the cookie loss became visible.

Fixed `hire-api.env` (backup `hire-api.env.bak-20260728-gmailfix`) and released. Live verification, the exact request from the report:

```
GET https://freehire.me/api/v1/me/gmail/callback?state=abc&code=xyz   (no cookie)
-> 302 Location: https://freehire.me/my/inbox?gmail_error=auth
GET https://freehire.dev/... same -> 302 to the same place
```

No JSON dead end on either host. **Superseded:** the "still required" steps 2 and 3 in the description — `SERVED_HOSTS` already lists all four hosts and sign-in is host-aware, so folding `.dev`'s `/api/` into the 301 is not needed for this bug and would sign out `.freehire.dev` sessions plus break API clients that were deliberately left on `.dev`.

Known remainder, not hit in practice: the connector still builds its `redirect_uri` once at startup instead of per request host, so a connect started on a non-canonical served host would still land cookie-less — now as an explained redirect rather than a dead end. Not reachable today since the `.dev` SPA is 301'd.

PR #1188, opened 2026-07-28
feat(sources): crawl the a16z talent network only where it isn't a mirror

strelov1:
## What

`speedrun-talent-network.com` (the a16z / Speedrun portfolio job board) now exposes a **public read-only JSON API** — `/api/v1/jobs`, `/companies`, `/collections`, per-role detail with the body, the comp band and the application target, plus `jobs.rss` and `openapi.json`. When it was harvested in July it was a Cloudflare-fronted Astro site with no API, and we deliberately shipped no adapter (#1070).

The API changed the cost, not the conclusion. A fresh coverage sweep — one apply-URL mined per company, 315 companies, matched against `sources/*.yml`:

| ATS behind the listing | companies | not in our board files |
|---|---:|---:|
| Ashby | 217 | 3 |
| Greenhouse | 63 | 0 |
| Lever | 22 | 1 |
| none — application hosted on the network | 13 | 13 |

**302 of 315 companies are already crawled natively.** Ingesting the network's 10,293 roles wholesale would be ~99% duplicates of boards we already hold. So this PR takes only the two slices that are genuinely new.

## The adapter — `internal/sources/speedrun.go`

`board` = the company slug. Listed **only** for the 13 early-stage companies that host their application on the network itself; their 16 roles reach us through no other source. All 13 are Speedrun-cohort companies — this is the early-discovery seam the July harvest called out: a cohort startup surfaces here before it adopts an ATS, and moves to that provider's board file when it does.

- Reads `/api/v1/companies/{slug}` for the live role ids, then each role's `/api/v1/jobs/{id}` for the body and apply URL. The `?company=` filter keys on the **display name** (`Brief`, not `brief`), which makes a poor board id — hence the company endpoint.
- `URL` is the application target: the company's own ATS when the role federates out, the network's page when it hosts it.
- `description_text` is plain text, so it goes through `plainTextToHTML` (blank lines → paragraphs, `- ` → `<ul>`) before `sanitizeHTML`; rendered raw it would collapse into one wall of text.
- The comp band leads the description, the way `hhru.salaryParagraph` does — `Job` has no salary field.
- `Seniority` and `Category` stay **unset on purpose**. The API's `function`/`seniority` labels are the network's own normalization of the title, not a field an employer filled in, and an adapter-set facet outranks the pipeline's dictionaries — a coarse `engineering` would overwrite a precise `backend`/`ml_ai` classification.
- A role the detail endpoint reports `status: "closed"`, or whose detail fetch fails, is dropped; an unreadable company page fails the board.

## The four missing native boards

Found by the sweep, each validated live against its own ATS API and through our own adapter: `Sui Foundation`, `walrus`, `XMTP Labs` (ashby) and `VeeFriends` (lever). Two carry a space in the board id — 36 existing ashby boards already do, and Go escapes the path.

## Verification

- `go test ./...` green; `go vet`, `gofmt` clean.
- Unit tests cover the two-phase fetch, the federated-vs-hosted apply URL, closed/unreachable roles, the employment-type enum, registry + facet registration, and that `sources/speedrun.yml` parses and validates (mutation-checked: a bogus provider in the file does fail it).
- Live smoke against the real API through the registry: all 13 boards resolve, 16 roles, every one with a non-empty description; all 4 native boards resolve through `ashby`/`lever`.

## Ops

`sources/speedrun.yml` is a new board file, so it needs its own ingest cron entry on prod (one schedule per file) — it is not picked up by any existing unit.

### DIFF
--- a/internal/sources/registry.go
+++ b/internal/sources/registry.go
@@ -161,6 +161,9 @@ func All(c HTTPClient) map[string]Source {
 		NewCrelate(c),
 		// Ashby boards whose public Posting API is disabled, served via the embed GraphQL.
 		NewAshbyGraphQL(c),
+		// VC talent network, board = portfolio company slug: listed only for the early-stage
+		// companies that host their application there instead of on an ATS of their own.
+		NewSpeedrun(c),
 		// Multi-company aggregators (boardless): one global feed, company per posting.
 		NewTecla(c),
 		NewTeamex(c),
--- a/internal/sources/speedrun.go
+++ b/internal/sources/speedrun.go
@@ -1,0 +1,148 @@
+package sources
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+// speedrunBaseURL is the a16z speedrun talent-network public API root.
+const speedrunBaseURL = "https://speedrun-talent-network.com/api/v1"
+
+// speedrunSourceTag identifies our crawler to the API, which asks every integration to name
+// itself so it can tell agent traffic apart.
+const speedrunSourceTag = "freehire"
+
+// speedrun adapts the a16z speedrun talent-network API. The network is mostly a federating
+// directory — a portfolio company's roles usually live on its own Ashby/Greenhouse/Lever board,
+// which we crawl natively — so the boards listed here are the early-stage companies that have
+// no ATS of their own yet and host their application on the network itself. Those roles exist
+// nowhere else.
+//
+// Board is the company slug. The company endpoint enumerates its live roles (the ?company=
+// filter takes the display name, which makes a poor board key), and each role's body, work
+// arrangement and application target come from its own detail endpoint.
+type speedrun struct {
+	http JSONGetter
+}
+
+// NewSpeedrun builds the speedrun talent-network adapter over the given HTTP client.
+func NewSpeedrun(c JSONGetter) Source { return speedrun{http: c} }
+
+func (speedrun) Provider() string { return "speedrun" }
+
+// speedrunCompany is a company page: the directory record plus the ids of its live roles. Only
+// the ids matter — every field we map comes from the role's own detail payload.
+type speedrunCompany struct {
+	Company struct {
+		Jobs []speedrunListItem `json:"jobs"`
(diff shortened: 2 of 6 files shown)


strelov1:
## Dupe audit against the live prod catalogue — three entries removed

The first pass matched mined apply-URL tokens against `sources/*.yml` only. Two holes in that method, both found by querying prod before merging:

**1. The mined tokens were never URL-decoded.** `Sui%20Foundation` was compared against the board file's `sui foundation` and reported missing. It is already there — `sources/ashby.yml:7586`, 3 open postings on prod. Entry removed. (It would not have duplicated rows: `dedupeBoards` collapses case-variant boards at load, since `NamespaceExternalID` uses the literal board string. It would just have logged a dropped board on every run.)

**2. "Applies on the network" ≠ "we don't have the company."** Two of the 13 hosted-application companies run a board we already crawl:

| Company | Already ingested as | Overlap |
|---|---|---|
| Brief | `briefhq` (its own careers page) | both roles, identical titles |
| Bead AI | `gem` / `bead-ai` | 4 open roles incl. the same Founding Account Executive |

Both removed from `sources/speedrun.yml`, and the file now carries a comment saying to check the catalogue by company, not by apply URL, before adding an entry.

**Verified clean** (prod query by company name, then by exact title): the remaining 11 companies — Cascade, ElectroKare, Infragrid, Lekondo, Maestro, Primitive Labs, Snag, Sourcerer AI, Taya, Tenera, URSA Mining — appear nowhere in the catalogue under any source. 13 net-new roles.

**Kept, deliberately:**
- ashby `walrus` (Walrus Foundation) — the existing `walrusfi` board is a *different* Ashby tenant with different postings (3 vs 1, no overlap).
- ashby `XMTP Labs` — prod has XMTP only through the `jobstash` aggregator.
- lever `VeeFriends` — 0 net-new postings, but it is the native source for a role we currently hold only as a `jobstash` mirror of that very Lever URL. Their main board (`rippling` / `veefriends-llc`, 10 roles) is unaffected.

Net: 13 new roles from the network + 2 new native boards, no duplicates.

PR #1248, opened 2026-07-29
feat(assistant): the agent is open to every signed-in user

strelov1:
## Why

`requireRollout` refused anyone who was neither a moderator nor a beta tester — on production that is **358 of 360 accounts**. But the product had already stopped acting restricted:

- `accountNav.ts` links "Agent" for everyone
- `/tailor` mounted the same chat with `requireBeta={false}`
- `AssistantChat.svelte`'s `requireBeta` prop defaulted to `false`, so no host ever turned the UI mirror on

What a non-member actually met was a **wall of 403s**. Opening `/my/assistant` fires three refused requests in one page load: `boot()` catches the refused session list, carries on with an empty array, and the emptiness is then read as "this user has no chats" — which the next branch answers by `POST`ing a new session, earning a second refusal.

The gate's stated reason had expired. It was written when inference was the user's to pay for; the comment beside it still claimed the agent "runs on the user's own machine with their own Claude subscription", which stopped being true when #1165 moved the agent in-process.

## What changed

- `requireRollout` is gone. Every `/api/v1/assistant/*` route keeps `mw.key` — cookie, session JWT, or full-scope API key — and nothing else.
- The dead UI mirror comes out: the `requireBeta` prop, the `allowed` derivation, the restricted-rollout notice, and `/tailor`'s `requireBeta={false}`.
- Stale comments corrected in `accountNav.ts` and `my/+layout.svelte` (the latter claimed the Agent entry was hidden from non-beta users; it never was).
- Docs updated: root `AGENTS.md`, `internal/handler/AGENTS.md`, `internal/assistant/AGENTS.md`.
- OpenSpec change `open-assistant-to-all-users` with the delta against `assistant-sessions`.

`beta_tester` stays on the user model and on `/auth/me` — this retires one consumer, not the mechanism.

## Tests

The two rollout tests are inverted rather than deleted — the contract deserves a test that fails if the gate comes back:

- `TestAssistantServesAUserInNoGroup` — a user in no group is served (201), **and** a caller with no credential is still refused (401). Both halves, because "open to everyone" is one edit away from "open to anyone".
- `TestAssistantServesABearerCallerInNoGroup` — same over Bearer.

Watched both fail (403, want 201) before removing the gate.

Green: `go build`, `go vet`, `go test ./...`, `gofmt`, `go test -tags=integration ./internal/handler/`, web `vitest` 447/447, `eslint` (0 errors), `pnpm build`.

## ⚠️ The debt this takes on

**Nothing meters an assistant turn.** Credits are charged in `cv_tailor.go`, `cv.go`, `match_analysis_stream.go`, `contributions.go` and `intake.go` — `internal/assistant` has no notion of them. The rollout gate was the only thing bounding that spend, and it is now gone.

Credit metering is an owed follow-up, filed separately. Until it lands, spend is watched rather than capped; re-closing the gate is a revert of this commit.

## Noted, not fixed

- `visibleAccountNav()` filters on `betaOnly` / `moderatorOnly` / `moderatorOrBeta`, but **no nav item carries any of those flags** — the whole mechanism is dead. Pre-existing, left alone.
- `boot()` still conflates "the list was refused" with "the list was empty". That conflation is what turned one 403 into three; with the refusal gone it has no trigger left in this path, so it is noted as a seam rather than built against a hypothetical future gate.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -102,7 +102,7 @@ Each is self-contained and can be read independently.
 - **Embeddings:** Queue-driven (`semantic_outbox`), incremental, reconciled by `reindex --semantic`
 - **Dictionaries:** All facet dictionaries are dict-only in production — never guess, emit nothing for unknowns
 - **Job deletion:** The lifecycle only soft-closes; `cmd/prune` is the sole hard-delete path
-- **In-app assistant:** a bounded tool-calling loop in-process (`internal/assistant`), streamed over SSE, gated to the restricted rollout. Tools act as the authenticated caller — no credential is minted for an agent
+- **In-app assistant:** a bounded tool-calling loop in-process (`internal/assistant`), streamed over SSE, open to every signed-in user. Tools act as the authenticated caller — no credential is minted for an agent. A turn is not metered against AI credits yet, so authentication is the only thing bounding the spend
 - **Experience provenance:** every banked achievement records whether the CANDIDATE asserted it (`cv_import`/`stated_in_chat`/`manual`) or the MODEL did (`agent_inferred`). Only the former may be written into a CV, and the check lives in the service path, not in a system prompt
 - **Sentry:** Opt-in, env-gated, errors-only — `sentry.Init` with `SendDefaultPII:false`
 - **Naming — "CV", not "résumé":** Default new surfaces to **CV**. Don't mass-rename the existing `resume`/`resumeextract` packages and columns — churn without value
--- a/internal/assistant/AGENTS.md
+++ b/internal/assistant/AGENTS.md
@@ -112,8 +112,10 @@ is never mistaken for the answer.
    the valid ones. That message is the model's only path to self-correction.
 
 ## Limitations
-- No metering. The assistant is free behind the restricted-rollout gate;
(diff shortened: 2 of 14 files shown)


strelov1:
Rebased onto main (5 commits behind, conflicting).

**The rebase surfaced something worth calling out:** main added a sixth assistant route — `POST /assistant/sessions/:id/autopilot` — and it was mounted behind the same `gate`. Opened it too, keeping its cookie-only carrier and its reason:

```go
api.Post("/assistant/sessions/:id/autopilot", mw.cookie, h.PostAssistantAutopilot)
```

Had the rebase been resolved by taking my side wholesale, that route would have vanished; had it been taken theirs, the gate would have survived in the newest feature. Both halves kept deliberately.

Second conflict was in `AssistantChat.svelte`, where main added the `openingActions` prop next to the `requireBeta` I removed — kept the former, dropped the latter.

Re-verified on the new base: `go build`, `go vet`, `gofmt`, `go test ./...`, `go test -tags=integration ./internal/handler/` (21s, ok), web vitest **457/457**, eslint 0 errors, `pnpm build` clean.

PR #1260, opened 2026-07-29
fix(tailor): an empty conversation shows the way in, and the active pane is legible

strelov1:
## What and why

Three things from using the workspace on production.

**1. A CV opened by id showed an empty chat with no way in.** I gated the two opening actions on \`resuming\` — "opened with \`?cv=\`" — but a CV re-opened by id can carry a conversation that was bound at bootstrap and never spoken to. That case rendered "Ask the agent anything to get started" and nothing else, which is indistinguishable from a chat whose history was lost. Reported as exactly that.

The chat component already renders the actions only while \`chat.messages.length === 0\`, so my flag was a second gate on the same question — and the two disagree in precisely the case that was reported. \`openingFor(resuming)\` is now \`openingActions()\`: the host says what it offers, the chat decides when.

**2. The active tab was not readable as active.** Editor / Settings / Chat marked the current one with \`bg-muted\`, which against the panel's own background is nearly invisible. Now the brand tint — the palette's olive (\`--brand: #5b6f00\`, \`--brand-muted\` as its soft fill, \`--brand-strong\` for text on it) — plus a heavier weight. Same treatment in the artifact panel, so both tab strips read alike. No new token.

**3. Navigation.** The account section is renamed **Tailor**: every CV it lists is aimed at one vacancy, and "CV builder" named the tool it grew out of. The header menu now repeats **Agent** and **Tailor** beside the **Inbox** it already carried — those three get opened in the middle of other work, so a trip through the account shell to reach them is a navigation for nothing. Icons come from the existing \`accountNavIcons\` map, so the menu and the rail cannot drift.

Route addresses are unchanged; only the label moved.

## Verification

\`svelte-check\` 0 errors, \`pnpm run lint\` clean, 467 web tests green, \`pnpm run build\` green. Go untouched and still green.

The unit tests follow the new contract: the actions are offered regardless of how the workspace was opened, they still send nothing by themselves, and the section's label is asserted so a rename cannot happen silently.

Not verified: the visual result on production — that is the next thing to look at, and 4.2 in the change's tasks is left open for it.

OpenSpec: \`openspec/changes/tailor-workspace-orientation/\`.

## Checklist

- [x] \`go build ./...\`, \`go vet ./...\`, and \`gofmt -l .\` (prints nothing) pass.
- [x] \`go test ./...\` passes (and \`go test -tags=integration ./...\` if I touched DB/handlers).
- [x] I regenerated committed artifacts when their source changed — none changed.
- [x] For \`design-system/\` changes: n/a.
- [x] For \`web/\` changes: lint, \`svelte-check\` and \`pnpm run build\` pass.
- [x] This stays within freehire's core/extension boundary (see CONTRIBUTING.md) — it does not bloat the core.

### DIFF
--- a/internal/cv/store.go
+++ b/internal/cv/store.go
@@ -77,6 +77,7 @@ type Repository interface {
 	SnapshotForAutopilot(ctx context.Context, id uuid.UUID, userID int64) (int64, error)
 	SetAutopilotReport(ctx context.Context, id uuid.UUID, userID int64, report []byte) (int64, error)
 	RevertAutopilot(ctx context.Context, id uuid.UUID, userID int64) (db.RevertCVAutopilotRow, error)
+	GetTailoredForJob(ctx context.Context, userID, jobID int64) (db.GetTailoredCVForJobRow, error)
 }
 
 // Seeder provides the user's extracted résumé, used to seed a base CV when they have none.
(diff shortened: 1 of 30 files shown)


strelov1:
Wrong base — this branch already merged as #1258, so the PR repeats its commits. Reopening from a clean branch off main.

PR #1282, opened 2026-07-30
fix(sitemap): bound what one sitemap request reads, and put the site schema back on the homepage

strelov1:
## Why

Probing the live site for an SEO audit turned up that the sitemap layer is **intermittently absent**, not slow. Inside one ten-minute window on 2026-07-29, `/sitemap.xml`, `/api/v1/companies/sitemap/boundaries` and a company chunk all returned **504** at nginx's 60s ceiling; twenty minutes later the same URLs answered in 0.95s / 3.8s / 0.9s. `/sitemap-jobs.xml` measured **57s** and **29s** on two probes minutes apart.

A 5xx on the index costs every sub-sitemap, because the index is the only URL `robots.txt` advertises — the healthy static-pages (133 URLs) and insights (100 URLs) files become undiscoverable with it.

Measured per layer before changing anything, so the fix targets the real cause:

| layer | result |
|---|---|
| boundary walk in psql | 2.8s |
| 50k company chunk in psql | 0.9s warm |
| same chunk **with** the hiring filter, no covering index | **8.4s** |
| API on `127.0.0.1:8082` (bypassing nginx) | 4.8s / 200 |
| through nginx locally | 5.4s / 200 |

So the SQL is not the bottleneck — buffer-cache contention with the ingest is, and the band is ~20×. The queries need to be *small enough that the slow end still fits*, which is what this PR does.

## Sitemap

- **Company chunk 50k → 10k.** The chunk is a latency budget, not a protocol allowance.
- **Hiring companies only** (`job_count > 0`), the same scope `/companies` lists. ~90k of ~299k rows are job-less reference imports (YC, company-info) whose pages have nothing to rank — they were spending crawl budget to say nothing.
- **`count(*) OVER ()` → a `max(slug)` probe.** The old query bought an exact last-row guard by materializing every row of the walk; `max(slug)` excludes exactly the same row in one index probe.
- **Migration 0057** — `companies_sitemap_hiring_idx (slug) INCLUDE (updated_at) WHERE job_count > 0`. Required, not an optimization: without it the hiring filter made a *smaller* 20k chunk **slower** (8.4s vs 0.9s), because `companies_pkey` carries neither the predicate nor `updated_at` and sent every candidate row to the 7.7 GB heap.
- **`jobSitemapFreshest` 25k → 15k** for the same margin. The freshest-jobs file keeps its deliberate partial coverage.
- **Six indexable pages added** to the static file, which nothing else linked a crawler to: `/open`, `/trends`, `/docs/api`, `/contribute`, `/status`, `/privacy`.

## SEO / GEO

- **The homepage emitted no JSON-LD at all.** `WebSite` + its `SearchAction` and the publisher `Organization` moved to `/about` when `HomeView` did, and `homeFaq.ts` still described them as living in `routes/+page.svelte` — a regression from that move, not a decision. They are back on `/`; `/about` keeps its copy plus the `FAQPage`, which has to stay where the visible FAQ is.
- **AI assistants' link fetchers were being redirected.** `Perplexity-User`, `Claude-User`, `ChatGPT-User` and `meta-externalagent` carry no `bot` in their UA, so `CRAWLER_UA` missed them and the first-visit redirect sent them to `?work_mode=remote&regions=global`. Those are exactly the fetches that get quoted in an answer, so the page an assistant cited was the filtered slice rather than the homepage.
- **`llms.txt`** gains `/open`, `/insights` and the feature landings, plus the catalogue's order of magnitude — rounded down and pointing at `/open` for live figures, since a static file shouldn't claim a precise number.

## Verification

- `go test -tags=integration ./internal/db/ ./internal/handler/` — the sitemap suites pass, incl. two new cases: a chunk-edge final row is still excluded, and a company with nothing open is listed nowhere.
- `go build ./...`, `go test ./...`, `gofmt`, `go vet` clean. Web: 511 unit tests, eslint on the touched files, `pnpm build`.
- The built SSR server run against the **live prod API** emits `WebSite`/`SearchAction`/`EntryPoint`/`Organization` in the homepage's initial HTML, and serves a 139-URL static file containing all six new paths.

## Deploy notes

- `companies_sitemap_hiring_idx` was pre-built on prod **CONCURRENTLY** ahead of this merge, so the migration's `IF NOT EXISTS` is a no-op and the release takes no lock on the constantly-upserted `companies` table.
- Follow-up, not in this PR: nginx has no `proxy_cache` for freehire. Copying the logo proxy's own `proxy_cache_use_stale error timeout updating http_50x` + `background_update` pattern onto `/sitemap*.xml` would make a slow window serve the previous sitemap instead of a 504. Bounded queries narrow the band; the cache is what removes the failure mode.
- Also observed, out of scope: `companies` carries a **7.7 GB main heap for 320k rows** — impossible under the 8 KB page limit, so 3-10× bloat (its PK index is 44 MB). `pg_repack` would speed up every companies read, not just these.

### DIFF
--- a/internal/db/companies.sql.go
+++ b/internal/db/companies.sql.go
@@ -43,17 +43,25 @@ func (q *Queries) CompanyJobCountBySlug(ctx context.Context, slug string) (int32
 const companySitemapBoundaries = `-- name: CompanySitemapBoundaries :many
 SELECT slug FROM (
   SELECT slug,
-         row_number() OVER (ORDER BY slug) AS rn,
-         count(*) OVER () AS total
+         row_number() OVER (ORDER BY slug) AS rn
   FROM companies
+  WHERE job_count > 0
 ) t
-WHERE rn % $1::bigint = 0 AND rn < total
+WHERE rn % $1::bigint = 0
+  AND slug < (SELECT max(slug) FROM companies WHERE job_count > 0)
 ORDER BY slug
 `
 
-// The slug ending every full chunk of `chunk_size` companies (ordered by slug),
-// excluding the final row, so the sitemap index can list each company sub-sitemap's
-// keyset cursor.
+// The slug ending every full chunk of `chunk_size` hiring companies (ordered by
+// slug), excluding the final row, so the sitemap index can list each company
+// sub-sitemap's keyset cursor. Same `job_count > 0` scope as ListCompanySitemap, or
+// the cursors would not line up with the chunks they open.
+//
+// The last-row guard is a max(slug) probe, not the `count(*) OVER ()` this query
+// used to compare `rn` against: that count is exact only by materializing every row
+// of the walk in a tuplestore, while max(slug) over the same partial index is one
+// backward index probe. Both exclude exactly the row whose rn = total — the maximum
+// slug — whose cursor would open an empty trailing chunk.
 func (q *Queries) CompanySitemapBoundaries(ctx context.Context, chunkSize int64) ([]string, error) {
 	rows, err := q.db.Query(ctx, companySitemapBoundaries, chunkSize)
 	if err != nil {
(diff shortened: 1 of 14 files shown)


strelov1:
reopening to trigger CI — the pull_request workflows did not fire on open

PR #1319, opened 2026-07-31
fix(handler): finish the fixture rename #1314 left half-done

strelov1:
`#1314` renamed `agentInboxFixture` → `harnessInboxFixture` together with the file that defines it (`inbox_agent_integration_test.go` → `inbox_harness_integration_test.go`), but `followup_integration_test.go` kept calling the old names — one type reference and six constructor calls.

`internal/handler` has not compiled under `-tags=integration` since that merged, so **the backend job is red on main and on every PR opened against it**, including #1317.

The new fixture already offers everything the followup test uses — `pool`, `q`, `userID`, `callKey` — so the fix is the rename and nothing else.

```
before: vet: followup_integration_test.go:19:38: undefined: agentInboxFixture
after:  go build -tags=integration ./... && go vet -tags=integration ./...  clean
```

Worth noting why nothing caught it: these files only compile under the `integration` build tag, and the unit job runs `go test ./...` without it. The rename passed every check that was looking.

### DIFF
--- a/internal/handler/followup_integration_test.go
+++ b/internal/handler/followup_integration_test.go
@@ -16,7 +16,7 @@ import (
 )
 
 // appliedDaysAgo backdates an application so it lands on the far side of its stage's threshold.
-func appliedDaysAgo(t *testing.T, f *agentInboxFixture, jobID int64, days int, stage string) {
+func appliedDaysAgo(t *testing.T, f *harnessInboxFixture, jobID int64, days int, stage string) {
 	t.Helper()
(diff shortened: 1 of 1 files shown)


strelov1:
Superseded by #1318, which landed the same seven-line rename first. Closing rather than rebasing — the diff would be empty.

Leaving the diagnosis here since it is the more useful half: `internal/handler` did not compile under `-tags=integration` between #1314 and #1318, so the backend job was red on main and on every PR opened in that window. Nothing caught the rename because these files only compile under the build tag, and the unit job runs `go test ./...` without it.

PR #1604, opened 2026-08-06
fix(assistant): close dangling tool_use when replaying history

millerscout:
## What and why

A mid-turn failure (cancelled request, transport error, process restart) can leave an assistant `tool_use` in the transcript without its matching `tool_result`. The next user message then fails at the provider — Bedrock returns 500 with `tool_use ids were found without tool_result blocks` — so the tailor/assistant session looks dead even though infra is fine.

`trim` already dropped *leading* orphan tool results after history windowing. This change also closes the inverse: before replay, synthesise an error `tool_result` for any unanswered `tool_use` so the conversation stays provider-legal and the session remains resumable (matching the package’s “partial transcript must be resumable” rule).

Closes #

## Checklist

- [x] `go build ./...`, `go vet ./...`, and `gofmt -l .` (prints nothing) pass. *(scoped to `./internal/assistant/` for this change)*
- [x] `go test ./...` passes (and `go test -tags=integration ./...` if I touched DB/handlers). *(unit tests for `./internal/assistant/`; no DB/handler signature change — `go vet -tags=integration ./internal/assistant/` clean)*
- [x] I regenerated committed artifacts when their source changed (`make sqlc` for SQL/migrations, `make gen-contracts` for contract types). *(n/a)*
- [x] For `design-system/` changes: `pnpm run check` and `pnpm run build` pass. *(n/a)*
- [x] For `web/` changes: `pnpm run check` and `pnpm run build` pass. *(n/a)*
- [x] This stays within freehire's core/extension boundary (see CONTRIBUTING.md) — it does not bloat the core.

## Test plan

- [ ] `go test ./internal/assistant/ -run TestHistoryClosesDanglingToolCallsBeforeReplay`
- [ ] Reproduce with a transcript that has assistant `tool_calls` followed by a user message and no tool result — next turn should succeed instead of provider 500
- [ ] Normal tool rounds (tool_use → tool_result → answer) unchanged

### DIFF
--- a/internal/assistant/AGENTS.md
+++ b/internal/assistant/AGENTS.md
@@ -189,7 +189,11 @@ it is the only tool whose completion depends on a client we do not control.
 **History trimming.** `trim` keeps the most recent N messages and then drops any
 leading tool results whose originating call was trimmed away — providers reject a
 tool result that answers no call in the conversation, so an orphan at the head
-turns a context loss into a failed turn.
+turns a context loss into a failed turn. It also closes the inverse shape: a
+`tool_use` whose `tool_result` never landed (a turn that died after persisting
+the calls). Replay synthesises an error result for each unanswered call so the
+next turn stays provider-legal — Bedrock in particular rejects the whole request
+otherwise.
 
 **Streaming.** `llm.Chat` splits the model's two delta channels: answer text goes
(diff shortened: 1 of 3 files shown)


strelov1:
👀

millerscout:
Thanks @strelov1 also, nice project!

PR #1717, opened 2026-08-10
feat(sources): add NoDesk source adapter

### LINKED ISSUE #1629: feat: add NoDesk source adapter
### Source kind

### Source identifier

NoDesk — nodesk — boardless (remote-jobs board)

### Sample listing/API URL

https://nodesk.co/remote-jobs/index.xml

### Notes

Public, zero-auth RSS/XML feed (same shape as our existing `weworkremotely` adapter — mirror it).

**Verified live (2026-08-08):** returns a standard RSS 2.0 feed. Each `<item>` has `title` ("Role at Company"), `description` (plain text body), `pubDate`, `guid`, `link`. Title needs splitting on " at " to separate role from company (same trick as WWR). All-remote board, so location is always "Remote".

robots.txt has no relevant disallow; the feed lives under `/remote-jobs/` which isn't blocked.

### Confirmation

HafizMMoaz:
## Summary
- Adds a boardless RSS adapter for NoDesk (`nodesk.co/remote-jobs/index.xml`), mirroring the existing `weworkremotely` adapter's shape (title split, all-remote board).
- The live feed embeds raw named entities (e.g. `&rsquo;`) outside CDATA, which the standard library's strict XML decoder rejects with "invalid character entity". Unlike `weworkremotely` (whose body is CDATA-wrapped, sidestepping this), NoDesk needed a lenient decode: fetched via `GetText`, decoded with `Strict = false` (leaves an unrecognized entity as literal text instead of erroring), then `html.UnescapeString` resolves it — the same helper `weworkremotely` already uses for its CDATA content.
- Registers the adapter in `sources.All` and adds `sources/nodesk.yml`.

Fixes #1629

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/sources/...` (includes a fixture reproducing the live feed's raw-entity shape, so a regression back to strict `GetXML` decoding fails loudly)

### DIFF
--- a/internal/sources/nodesk.go
+++ b/internal/sources/nodesk.go
@@ -1,0 +1,92 @@
+package sources
+
+import (
+	"context"
+	"encoding/xml"
+	"fmt"
+	"strings"
+	"time"
+
+	xhtml "golang.org/x/net/html"
+)
+
+// nodesk adapts nodesk.co, a remote-jobs board. Boardless (one public RSS feed, no
+// per-tenant board) and multi-company, so it stays in the source facet and takes each
+// posting's company from the feed. All-remote board, so every posting is remote. The feed
+// carries every posting's body inline (no detail call).
+//
+// Fetched via GetText rather than GetXML: unlike weworkremotely's feed (which wraps its body
+// in CDATA, leaving embedded entities as literal text), NoDesk's <description> contains raw,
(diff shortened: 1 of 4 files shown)


HafizMMoaz:

strelov1:
> @strelov1 ready for review whenever you get a chance.

Thank you!

PR #1718, opened 2026-08-10
feat(sources): add Cryptocurrency Jobs source adapter

### LINKED ISSUE #1632: feat: add cryptocurrencyjobs source adapter
### Source kind

### Source identifier

Cryptocurrency Jobs — cryptocurrencyjobs — boardless (Web3/crypto niche, all-remote)

### Sample listing/API URL

https://cryptocurrencyjobs.co/index.xml

### Notes

Public, zero-auth RSS/XML feed. All-remote board (100% of listings), same shape as `weworkremotely`/`nodesk`.

**Verified live (2026-08-08):** standard RSS 2.0. `<item>` has `title` ("Role, Product at Company"), `description` (plain-text summary), `pubDate`, `guid`, `link`.

robots.txt has a Content-Signal block (`ai-train=no, use=reference`) and disallows a handful of AI-training-specific crawlers (GPTBot, ClaudeBot, CCBot, etc.) by name, plus `Disallow: /*?` for wildcard query strings — none of that touches the plain `/index.xml` feed URL, which carries no query string and isn't in any disallow list. Worth using a plain, honestly-identified user agent rather than impersonating a browser, given the site's explicit stance on bot traffic.

### Confirmation

HafizMMoaz:
## Summary
- Adds a boardless RSS adapter for Cryptocurrency Jobs (`cryptocurrencyjobs.co/index.xml`), all-remote Web3/crypto niche board.
- Same feed shape as `weworkremotely`/`nodesk` ("Role at Company" title, plain-text description, guid = link). Uses the same lenient-decode approach just added for `nodesk` (#1629): fetched via `GetText`, decoded with `Strict = false`, then `html.UnescapeString` resolves any raw named entities the feed embeds outside CDATA — this board runs the same underlying feed generator family as nodesk.co (identical item shape), so the same non-standard-entity risk applies even though today's snapshot happened not to hit it.
- Registers the adapter in `sources.All` and adds `sources/cryptocurrencyjobs.yml`.
- robots.txt disallows a handful of named AI-training crawlers plus wildcard query strings, but the plain `/index.xml` fetch this adapter makes isn't in any disallow list.

Fixes #1632

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/sources/...`

### DIFF
--- a/internal/sources/cryptocurrencyjobs.go
+++ b/internal/sources/cryptocurrencyjobs.go
@@ -1,0 +1,91 @@
+package sources
+
+import (
+	"context"
+	"encoding/xml"
+	"fmt"
+	"strings"
+	"time"
+
+	xhtml "golang.org/x/net/html"
+)
+
+// cryptocurrencyjobs adapts cryptocurrencyjobs.co, a Web3/crypto-niche remote-jobs board.
+// Boardless (one public RSS feed, no per-tenant board) and multi-company, so it stays in the
+// source facet and takes each posting's company from the feed. All-remote board, so every
+// posting is remote. The feed carries every posting's body inline (no detail call).
+//
+// Fetched via GetText rather than GetXML and decoded leniently (Strict=false), same as the
+// nodesk adapter: this board runs the same underlying feed generator (identical "Role at
+// Company" / plain-text-description / guid-equals-link shape, same host family as nodesk.co),
+// which is known to embed raw named entities such as "&rsquo;" outside CDATA — invalid XML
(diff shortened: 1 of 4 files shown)


HafizMMoaz:

strelov1:
Thanks for the clean adapter — merged. One follow-up: the board isn't actually 100% remote (e.g. "Product Designer (New York only) at Loopscale" requires working from their NYC office), so hardcoding `Remote: true` / `Location: "Remote"` mislabels onsite postings. Filing a follow-up issue to fix the location parsing.

PR #1719, opened 2026-08-10
feat(sources): add Remotli source adapter

### LINKED ISSUE #1631: feat: add Remotli source adapter
### Source kind

### Source identifier

Remotli — remotli — boardless (remotli.ch, Swiss remote-jobs board)

### Sample listing/API URL

https://remotli.ch/api/jobs

### Notes

Public, no-auth JSON API. Boardless + aggregator shape, paginated.

**Verified live (2026-08-08):** returns `{jobs: [...], pagination: {page, limit, total, totalPages}}`. `limit=20`, `total=393`, `totalPages=20` at check time. Each element of `jobs[]` is itself wrapped one level deep: `{"jobs": {id, title, company, location, allLocations, type, salary, salaryMin, salaryMax, salaryCurrency, description, status, publishedAt, ...}}` — unwrap that inner key when mapping.

Small niche board (Swiss companies, roles paid in CHF), but the payload is unusually complete: `status` gives liveness at source (only emit `active`), `description` ships full in the list (no per-job detail request needed), `publishedAt` gives real recency, `company` is the real employer (not the board name).

**Caveat — robots.txt disallows `/api/`** for every listed user agent (`*`, GPTBot, ClaudeBot, PerplexityBot, etc.):
```
User-Agent: *
Allow: /
Disallow: /api/
```
Same situation as Landing.jobs in this batch — flagging so it's a conscious call, not a silent one.

### Confirmation

HafizMMoaz:
## Summary
- Adds a boardless JSON adapter for Remotli (`remotli.ch/api/jobs`), a Swiss remote-jobs aggregator. Paginated by `?page=N` over a reported `pagination.totalPages`; each posting is wrapped one level deep under its own `jobs` key, matching the live-verified shape from the issue.
- Only `status: "active"` postings are emitted — the field gives liveness at source.
- `Remote` is set from the shared `isRemote()` free-text location heuristic rather than a structured flag: unlike this batch's other RSS boards, Remotli mixes onsite Swiss postings (`"Zürich, Switzerland"`) in with remote ones, and the API's `remotePolicy` field was null on every sampled posting, so `WorkMode` is left empty for the pipeline's own location dictionary to derive.
- `EmploymentType` reuses the existing `schemaEmploymentType` mapper (Remotli's `type` uses hyphens — `"full-time"` — where the schema.org enum uses underscores, so the hyphens are normalized before the shared mapper runs).
- Registers the adapter in `sources.All` and adds `sources/remotli.yml`.

**Conscious call, not a silent one** (as the issue itself asked to flag): `robots.txt` disallows `/api/` for every listed user agent. The endpoint is public and unauthenticated, and the crawl is a scheduled job rather than a search-engine crawler, so this PR implements the adapter — noted in its doc comment for visibility.

Fixes #1631

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/sources/...` (pagination across two pages, active-status filter, onsite-vs-remote heuristic, first-page-hard-fail vs later-page-partial-result)

### DIFF
--- a/internal/sources/registry.go
+++ b/internal/sources/registry.go
@@ -222,6 +222,7 @@ func All(c HTTPClient) map[string]Source {
 		NewPowerToFly(c),
 		NewHimalayas(c),
 		NewRemotive(c),
+		NewRemotli(c),
 		NewTheMuse(c),
 		NewJustJoin(c),
 		NewNoFluffJobs(c),
--- a/internal/sources/remotli.go
+++ b/internal/sources/remotli.go
@@ -22,6 +22,9 @@ type remotli struct {
 
 const remotliListURL = "https://remotli.ch/api/jobs?page=%d"
 
+// remotliMaxPages bounds pagination so a wrong or missing totalPages cannot loop.
+const remotliMaxPages = 100
+
 // NewRemotli builds the Remotli adapter over the given HTTP client.
 func NewRemotli(c JSONGetter) Source { return remotli{http: c} }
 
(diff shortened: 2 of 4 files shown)


HafizMMoaz:

PR #1721, opened 2026-08-10
docs(assistant): note REALTIME_MODEL needs the provider prefix

strelov1:
Small follow-up to #1716. Discovered while wiring the prod litellm proxy: `REALTIME_MODEL` needs the provider prefix (`openai/gpt-realtime-2.1`), not the bare alias — the litellm proxy's realtime WebRTC feature doesn't resolve a `model_name` alias the way the rest of the proxy does. Doc-only, no code change. Prod is already configured correctly; this just writes it down.


strelov1:
Superseded by a clean branch — this one carried the pre-squash history from #1716 and GitHub flagged it conflicting. Same one-line doc change, rebased cleanly onto main.

PR #1724, opened 2026-08-10
feat(sources): add SolidJobs source adapter (IT division)

### LINKED ISSUE #1630: feat: add SolidJobs source adapter
### Source kind

### Source identifier

SolidJobs — solidjobs — boardless, per-division (Polish market)

### Sample listing/API URL

https://solid.jobs/public-api/offers/it?campaign=freehire

### Notes

Public JSON API, but **requires a `campaign` query param** — a bare `/public-api/offers/it` 400s with `"Make sure that campaign parameter exists and contains only lowercase letters, numbers and dashes (max 64 chars long)."`. Any lowercase-alphanumeric-dash value works; the value is echoed back into each returned job's `url` (`/o/<key>/<campaign>`), so pick a stable identifying value rather than a throwaway string.

**Verified live (2026-08-08)** with `?campaign=freehire`: 200, large payload (~2.4 MB for the `it` division alone). Item shape: `jobOfferKey`, `title`, `division`, `category`, `subCategory`, `company`, `companyLogoUrl`, `salary: {from, to, currency, period, employmentType}`, `contractTime`, `locations[]`, `isRemote`, `isHybrid`, `url`, `experienceLevel`, `skills[]`.

The board spans multiple divisions (`it` is one of at least 8 — the others weren't enumerated here, worth discovering their slugs before onboarding). One board-file entry per division, similar to how `whatjobs` boards are per-keyword.

robots.txt only disallows `/management/` and `/admin/` — the API path is unrestricted.

### Confirmation

HafizMMoaz:
## Summary
- Adds a board-based (not boardless) aggregator adapter for SolidJobs (`solid.jobs/public-api/offers/<division>`), a Polish job board split into divisions. Same keyword-as-board shape as `whatjobs`: the board-file entry's `board` carries a division slug rather than a tenant id.
- Maps `company`, `locations[]`, `description` (full HTML inline, no detail call), `isRemote`/`isHybrid` (structured → `Remote`/`WorkMode` via the existing `workModeFromRemoteHybrid`), `contractTime` → `EmploymentType`, `experienceLevel` → `Seniority` (the platform's own taxonomy, mapped through a small dict — `"Regular"` is the Polish market's usual label for mid-level), and `skills[].name` → `Skills` via the existing `skilltag.Canonicalize`.
- Registers the adapter in `sources.All` and adds `sources/solidjobs.yml` with one entry for the `it` division. The other ~7 divisions mentioned in the tracking issue were never enumerated live, so they're left for a follow-up rather than guessed.
- Requires a `campaign` query param (any lowercase-alphanumeric-dash string); used a stable `"freehire"` value since it's echoed back into every posting's own returned URL.

**Known limitation, verified live (2026-08-10) — flagging for visibility:** the response reports `totalPages`/`totalCount` (3 pages, 1399 jobs for `it`), but I could not find any way to reach a page past the first: `pageIndex`, `pageNumber`, `page`, `skip`, `offset` as query params, and a POST body, all silently return page 0 again. So only the first fixed 500 postings per division are reachable through this adapter today. Documented in the adapter's doc comment as a real gap for whoever finds the actual paging mechanism — real, deduped, live data either way, just a partial slice of the division rather than all of it.

Fixes #1630

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/sources/...` (field mapping, remote/hybrid work-mode, seniority + employment-type dictionaries, skill canonicalization, blank-board rejection)

### DIFF
--- a/internal/sources/registry.go
+++ b/internal/sources/registry.go
@@ -251,6 +251,9 @@ func All(c HTTPClient) map[string]Source {
 		// hh.ru: multi-company aggregator, enumerated by professional_role (board), reading the
 		// server-rendered search page's embedded state.
 		NewHH(c),
+		// SolidJobs: Polish job board, multi-company aggregator enumerated by division (board),
+		// same keyword-as-board shape as whatjobs.
+		NewSolidJobs(c),
 		// RU-domestic single-company adapters (boardless, except Yandex which selects
 		// host+language by board).
 		NewYandex(c),
--- a/internal/sources/solidjobs.go
+++ b/internal/sources/solidjobs.go
@@ -1,0 +1,143 @@
+package sources
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/strelov1/freehire/internal/skilltag"
+)
+
+// solidjobs adapts solid.jobs, a Polish job board split into divisions (IT, finance, ...).
+// Not boardless: like whatjobs, each board-file entry's board is a SEARCH SLICE rather than a
+// tenant id — here a division slug (e.g. "it") rather than a keyword — so it stays an
+// aggregator without the boardless marker. company is a display label only; each job's real
+// employer comes from the posting.
+//
+// The public endpoint requires a campaign query param (any lowercase-alphanumeric-dash value;
+// it is echoed back into every posting's own URL, so a stable identifier is used rather than a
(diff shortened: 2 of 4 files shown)


HafizMMoaz:

PR #1885, opened 2026-08-13
docs(design): tracker application-process aggregates — design pass

### LINKED ISSUE #1625: Design: anonymous sharing of tracker application-process data on job postings
## Idea

Users track their application process for a job (stages, dates, outcome) via the tracking/calendar feature (`internal/userjob`). Let users optionally share that process data — anonymized — so other users viewing the same job posting can see aggregate signal like:

- How many people applied / are tracking this job
- Typical time between stages (applied → screen → interview → offer/reject)
- Rejection rate / outcome distribution
- Maybe: time-to-first-response

This turns freehire's tracking data into a crowdsourced signal on the job itself — similar to what Blind/Levels.fyi/Glassdoor do for comp, but for process timing and odds, sourced from freehire's own users instead of scraped data.

## Why this matters

We already have the raw data: `internal/userjob` stages, `internal/tracking-calendar` events (`hire-tracking-calendar.md`), and the event ledger (`hire-application-event-ledger.md`). No new ingestion needed — this is an aggregation + presentation layer on data we already collect, opt-in and per-user.

## Needs a design pass before implementation (open questions)

- **Anonymity guarantee**: what's the k-anonymity floor (e.g. don't show aggregates until N ≥ 5 trackers) so a job with 1-2 trackers doesn't become identifiable?
- **Opt-in vs opt-out**: default off, with explicit consent per tracked job, or a one-time account-level setting?
- **What's shown**: raw counts vs buckets/ranges (e.g. "5-10 applicants" instead of exact numbers) to reduce inference risk?
- **Where it's not applicable**: jobs with very low tracker counts, or where `internal/userjob` events are sparse/inconsistent (users don't always log every stage)?
- **Staleness**: aggregates need to reflect the *current* job (reposts, is_repost clustering per `hire-repost-marker-lag-window.md`) not stale historical postings.
- **Abuse/gaming**: could a company or bad actor pollute the signal (fake tracked applications) to make a job look more/less competitive?

## Related existing docs

- `internal/userjob/AGENTS.md` — tracking stages, /me/tracking
- `hire-tracking-calendar.md` — day = reader's timezone, iCalUID-only linking
- `hire-application-event-ledger.md` — event ledger, CTEs blind to each other

Needs a brainstorming/design pass (product + privacy tradeoffs) before any implementation.

timon0305:
Design pass for #1625. **Design only — no implementation proposed.** Filed under `docs/superpowers/specs/` rather than as an openspec change, since that is where this repo puts a design pass awaiting approval and `openspec/changes/` presupposes the change.

## The framing needed correcting first

freehire already ships this one level up. `company-hiring-signal` maintains a per-company response rate and median time-to-first-reply over the same ledger, and it already answers most of what the issue lists as open:

| Issue's open question | Already decided in `company-hiring-signal` |
|---|---|
| Sample floor | **10** tracked applications — below it the field is *absent*, not zero |
| Which rows count | Only owners with a **connected mailbox**, on both sides of every ratio |
| Source | `application_events`, never live `emails` — inbox hygiene must not move a public rate |
| Censoring | Unanswered excluded from the median, reported as a count beside it |

So the real question is whether that design survives being pushed down to a single posting.

## It does not — and a bigger k-floor does not save it

1. **The adversary knows the input set.** A company aggregate pools across postings and years. A *posting* aggregate is computed over the exact list of applicants the employer already holds in their own ATS. k-anonymity assumes an adversary who cannot enumerate the population; here they define it.

2. **A live counter leaks by differencing.** Reject one named candidate on Tuesday; if `rejected` moves 2→3 on Wednesday, that person is identified as a tracker and their outcome confirmed. A publication floor gates *whether the value is shown*, not *how it moves once shown* — so N ≥ 5, or 50, changes nothing. This is the argument the issue's proposed floor misses, and it is what decides the design.

3. **A median over five observations is one person's value** — the third one. Timing needs a much higher floor than counts, and the issue treats them as one question.

4. **Stage timing is largely not computable today, deliberately.** `stage_set` is written from `user`/`assistant`/`system` sources and **none** is admitted by `appevent.TrustedForDayMath`, because a hand-typed stage measures the candidate's diligence rather than the employer's behaviour. The ledger spec already has a scenario called *"Stage velocity starts empty and honest"*. The issue's second bullet is therefore not waiting on an aggregation layer — its inputs are untrusted by design.

## Recommendation

Do not build at posting grain. Aggregate at the **role cluster** — already a first-class key, indexed by `jobs_company_role_fingerprint_idx` since migration `0003` to back the reality signal's own per-cluster counts. Reuse the existing **10**-application gate rather than invent a third number (two already exist: 10 here, `ghost.ContributorGate = 2` for distinct people). Count **DISTINCT users**, not applications — `user_jobs`' `(user_id, job_id)` PK is per *posting*, so one diligent tracker could otherwise clear the gate alone across a cluster's cities. Defend against differencing with **bucket hysteresis + a cohort freeze**, not a floor.

Also resolved in the doc: account-level opt-in (per-job consent manufactures selection bias — people share offers and withhold rejections), and abuse (the mailbox gate blocks fabricated replies; what remains reachable is denominator pollution, which *deflates* a rate).

## Run this before scheduling any of it

The doc carries a runnable read-only query for the distribution of distinct observable trackers per role cluster, with the `observable` CTE lifted verbatim from `RebuildInsightsCompanyResponse` so the measurement is taken under the exact gate the feature would ship with. If few clusters clear 10, the right outcome is to close #1625 as premature — a query, not a project.

## Notes for the reviewer

- The measurement query joins on `application_events.job_id`, which `cmd/prune` sets to NULL, so it slightly **under**-counts. Right direction for a go/no-go, but flagged in the doc as the reason it cannot double as the feature's own aggregation.
- Bucket boundaries, hysteresis width and the freeze window are deliberately left open — choosing them before the distribution is known would be a guess presented as a threshold.
- Docs-only: no Go, no schema, nothing generated. Every referenced path verified to exist.

## Summary by CodeRabbit

* **Documentation**
  * Added a design specification for privacy-preserving aggregate tracker insights.
  * Defines safeguards including role-level aggregation, distinct-user counting, minimum thresholds, bucketed reporting, fixed observation windows, and default-off account consent.
  * Documents deferred reporting capabilities, measurement guidance, data-retention considerations, and open questions for future development.

### DIFF
--- a/docs/superpowers/specs/2026-08-13-tracker-process-aggregates-design.md
+++ b/docs/superpowers/specs/2026-08-13-tracker-process-aggregates-design.md
@@ -150,6 +150,13 @@ informative, since it says somebody who was in this cluster left. The immutable-
 *What is shown*, below, absorbs it: a published window is never recomputed, so a revocation takes
 effect from the next window rather than retracting a number already shown.
 
+The same rule bounds the retroactivity above, and the two resolutions have to be read together:
+consent reaches back over a candidate's existing applications, but only as far as **windows that
+are still open**. It cannot enter a window already published, because that would let a consent
+change move a number already shown — the same door §2 closes, approached from the other side.
+Retroactive therefore means "every application whose window has not yet closed", not "everything
+they have ever done", and the difference is worth stating to whoever writes the consent copy.
+
 **What is shown.** Buckets, never raw counts, and buckets with **hysteresis** — a band must be
 crossed by a margin before the display changes, so no single arrival or state change can move
 what is rendered.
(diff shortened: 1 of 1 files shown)


strelov1:
Ran the read-only measurement query from this design against prod (host-2, 2026-08-13):

```
 trackers | cluster_count | companies
----------+---------------+-----------
        1 |           263 |       206
(1 row)
```

That's the entire result set — no `(company_slug, role_fingerprint)` cluster has more than 1 distinct observable tracker (mailbox-connected user with an `applied` event). Not even the `replySampleGate = 5` clears anywhere, let alone `responseSampleGate = 10`. And per the doc, this number is an *upper bound* (no opt-in predicate applied yet), so the real, consent-gated figure is only smaller.

Per the doc's own decision rule: a disappointing result here settles a no-go. Recommend closing #1625 as premature rather than proceeding with any of the design work — worth revisiting once the tracker base has grown enough for clusters to clear the gate.

timon0305:
> Ran the read-only measurement query from this design against prod (host-2, 2026-08-13):
> 
> ```
>  trackers | cluster_count | companies
> ----------+---------------+-----------
>         1 |           263 |       206
> (1 row)
> ```
> 
> That's the entire result set — no `(company_slug, role_fingerprint)` cluster has more than 1 distinct observable tracker (mailbox-connected user with an `applied` event). Not even the `replySampleGate = 5` clears anywhere, let alone `responseSampleGate = 10`. And per the doc, this number is an _upper bound_ (no opt-in predicate applied yet), so the real, consent-gated figure is only smaller.
> 
> Per the doc's own decision rule: a disappointing result here settles a no-go. Recommend closing #1625 as premature rather than proceeding with any of the design work — worth revisiting once the tracker base has grown enough for clusters to clear the gate.

Agreed, clear no-go. Fine  to close #1625 .
One thing: I didn't run the second query which is what the doc says confirms the upper bound. With max = 1 it's 9 short of the gate so it almost certainly doesn't matter, just nothing it. 
Also, the thing that needs to grow isn't user count, it's two people tracking the same role at the same company. 263 clusters across 206 companies is about as spread as it gets, and more postings spreads it further. So "re-run the query sometime" is probably a better trigger than a headcount.
I'd merge this doc rather than drop it with the issue, saves the next person starting over. Happy to update it with your result if useful.

PR #1886, opened 2026-08-13
feat(sources): add the Landing.jobs adapter

### LINKED ISSUE #1627: feat: add Landing.jobs source adapter
### Source kind

### Source identifier

Landing.jobs — landingjobs — boardless (EU tech job-board aggregator)

### Sample listing/API URL

https://landing.jobs/api/v1/jobs

### Notes

Public, no-auth JSON API, single call returns a top-level array (like `remotive`/`getmanfred` — boardless + aggregator shape).

**Verified live (2026-08-08):**
- Returns 50 items. `?page=N` appears to be ignored — identical body across `page=1` and `page=2` (byte-for-byte). Needs checking during implementation whether this 50 truly is the full active catalogue or whether a different param controls page size/depth — don't assume 50 is everything without confirming.
- Field shape per item: `title`, `url`, `locations` (array of `{city, country_code}`, can be `null` for fully-remote roles), `remote` (bool), `published_at` / `created_at` (ISO8601 with ms), plus rich HTML body fields (`role_description`, `main_requirements`, `nice_to_have`, `perks`), `tags`, `type`, `gross_salary_low/high`, `currency_code`.
- **No company-name field in the payload.** Postings are `https://landing.jobs/at/<company-slug>/<job-slug>`; company must be derived from the URL slug (humanize: hyphens → spaces, title-case).
- Location: take first entry of `locations[]`, format as "city, country_code"; append "Remote" when `remote` is true; handle `locations: null`.
- postedAt: `published_at`, fallback `created_at`.

**Caveat — robots.txt disallows `/api/`:**
```
User-agent: *
Disallow: /api/
```
Flagging this explicitly since it's the exact path this adapter would hit. Every other candidate I checked in this batch (except Remotli) has no such restriction. Worth a conscious call before merging, not a silent decision either way.

### Confirmation

timon0305:
Closes #1627. Boardless multi-company adapter over landing.jobs' public keyless JSON API, in the shape `remotive`/`getmanfred` already use.

## ⚠️ The robots.txt call is the thing to decide

`landing.jobs/robots.txt` disallows `/api/` — the exact path this adapter hits. The issue asked for **a conscious call before merging, not a silent decision either way**, so this PR does not make it quietly.

**This project has already made the same call once.** `internal/sources/remotli.go` carries this note, for a board whose robots.txt disallows `/api/` identically:

> the endpoint is public and unauthenticated, and the crawl is a scheduled job rather than a search-engine crawler, but this is flagged here because it was flagged in the adapter's tracking issue as a conscious call rather than a silent one

The adapter reproduces that reasoning in its own doc comment, in the same terms and naming remotli as the precedent. Merging is where the call gets ratified; reverting is one file plus one registry line if you'd rather not.

For completeness, the Lever case is **not** a counter-precedent — `cmd/harvest-boards` skips Lever because *Common Crawl's* bot is disallowed, so the corpus doesn't exist to discover from. That's inability, not a policy of respecting robots.

**I did not fetch the endpoint from my machine** — verifying the pagination question would have meant performing the disallowed access before you'd decided, which is the silent decision the issue rules out. Everything below is built to the issue's own live findings, and the one open question is handled defensively rather than guessed.

## Two shapes that make it unlike its neighbours

**No company and no id in the payload.** Both come from the URL, `landing.jobs/at/<company-slug>/<job-slug>` — employer from the first segment humanized, external id from *both* segments joined. Both, because a job slug alone isn't unique across employers and the id is the dedup key. A URL without that shape is skipped rather than ingested under a fabricated identity, which would re-insert on every crawl.

**One request, no pagination.** `?page=N` is ignored — pages 1 and 2 come back byte-identical per your check — so a page walk would re-fetch the same postings until its cap and dedup them all away, paying N requests for one page of data. A test pins the single request so nobody restores a walk without evidence a depth parameter exists. Whether the ~50 items are the whole active catalogue is **still unconfirmed**, and the adapter says so at the point where the walk would go.

## Deliberately conservative where the payload is unverified

`tags`, `type` and the salary trio are **undeclared**. Salary has no home on `Job` (enrichment owns it); the element shape of `tags` and the vocabulary of `type` were not verified live, and a wrong Go type for either fails the decode of the **whole feed**, not one field. Undeclared fields are ignored by `encoding/json`, so the safe move is to omit them and let the pipeline's dictionaries mine the description — with a comment saying exactly what it takes to add them once confirmed.

Same reasoning on `remote=false`: it states "not flagged remote", not onsite, and the board carries hybrid roles — so it sets no `WorkMode` and leaves the pipeline to derive one. `remote=true` does set it, since that direction *is* stated.

`Countries` comes from `locations[].country_code`, a structured field, via the existing `countryFromCode`.

## Verification

`gofmt -l` clean · `go vet ./internal/sources/` clean (also with `-tags=integration`) · `go test ./internal/sources/` passes · `go run ./cmd/validate-sources` → OK, 204 files.

12 new tests cover the URL-derived identity, the humanizer, query/fragment stripping, `locations: null`, the no-onsite-assertion rule, skip-on-unresolvable-URL, single-request, and list failure as a board-level error.

No `docs/sources.md` entry — it carries live catalogue counts and is refreshed from prod separately (remotli isn't in it either).

Not run here: `go build ./...` and repo-wide `gofmt -l .`, since this was worked on Windows where `cmd/reindex` needs Unix-only `syscall.Statfs` and `core.autocrlf` flags every file. Both pre-existing platform limits; the touched package was vetted and tested directly.

## Summary by CodeRabbit

- **New Features**
  - Added job listings from Landing.jobs.
  - Imported posting details including titles, companies, locations, countries, remote status, employment types, dates, and descriptions.
  - Improved country handling by removing duplicates and unresolved entries while preserving order.
- **Bug Fixes**
  - Excluded incomplete listings without a valid URL, ID, or title.
- **Configuration**
  - Registered Landing.jobs as an available job source.

### DIFF
--- a/internal/sources/helpers.go
+++ b/internal/sources/helpers.go
@@ -135,6 +135,29 @@ func countryFromCode(code string) []string {
 	return []string{c}
 }
 
+// countriesFromCodes is the multi-place counterpart of countryFromCode: it normalizes a
+// posting's whole set of ATS-supplied country codes, dropping unresolved and duplicate ones and
+// keeping first-seen order. A board that states several locations for one posting states several
+// countries (landing.jobs lists a single role across Munich, Lisbon and Cologne), and keeping
+// only the first would hide that posting from a filter on any of the others. Returns nil when
+// nothing resolves, so an adapter can wire it straight into Job.Countries.
+func countriesFromCodes(codes []string) []string {
+	var out []string
+	seen := make(map[string]struct{}, len(codes))
+	for _, code := range codes {
+		c := location.NormalizeCountry(code)
+		if c == "" {
+			continue
+		}
+		if _, dup := seen[c]; dup {
+			continue
+		}
+		seen[c] = struct{}{}
+		out = append(out, c)
+	}
+	return out
+}
+
 // distinctJoin maps each item to a label, drops blank and duplicate labels (keeping first-seen
 // order), and joins the rest with sep. Adapters that build a location from a list of place
 // objects share it (getmatch, habrcareer) instead of each re-looping.
--- a/internal/sources/landingjobs.go
+++ b/internal/sources/landingjobs.go
@@ -1,0 +1,217 @@
+package sources
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"unicode"
+)
+
+// landingjobs adapts landing.jobs, an EU tech job board. Boardless (one public API, no
+// per-tenant board) and multi-company, so it stays in the source facet and takes each
+// posting's employer from the posting rather than from the config entry.
+//
+// Two shapes make it unlike its neighbours.
+//
+// **Neither the company nor a posting id is in the payload.** Both are derived from the
+// posting URL, which is `https://landing.jobs/at/<company-slug>/<job-slug>`: the employer from
+// the first segment, humanized, and the external id from both segments joined — the only stable
+// identity the feed offers. A posting whose URL does not have that shape is skipped rather than
+// ingested under a guessed identity, since the id is the dedup key.
+//
+// **One request, and no pagination.** `?page=N` is ignored by the endpoint — pages 1 and 2 come
+// back byte-for-byte identical (verified 2026-08-08, issue #1627) — so a page walk would
+// re-fetch the same postings until its own cap and dedup them all away, paying N requests for
+// one page of data. The single response is therefore taken as the whole feed. Whether the ~50
+// items it returns ARE the whole active catalogue is unconfirmed; if a parameter controlling
+// depth is found, this is where the walk goes, and `internal/sources/AGENTS.md`'s first-page
+// rule applies to it.
+//
+// Note: landing.jobs' robots.txt disallows /api/ for every listed user agent. The endpoint is
+// public and unauthenticated, and the crawl is a scheduled job rather than a search-engine
+// crawler, but this is recorded here — as it is on remotli, the other adapter in this position —
+// because it was raised in the adapter's tracking issue as a conscious call rather than a silent
+// one.
+type landingjobs struct {
+	http JSONGetter
+}
+
+const landingjobsListURL = "https://landing.jobs/api/v1/jobs"
+
+// landingjobsPathMarker precedes the "<company-slug>/<job-slug>" pair in a posting URL.
+const landingjobsPathMarker = "/at/"
+
+// NewLandingJobs builds the landing.jobs adapter over the given HTTP client.
+func NewLandingJobs(c JSONGetter) Source { return landingjobs{http: c} }
+
+func (landingjobs) Provider() string { return "landingjobs" }
+
+// landing.jobs is one global feed, so its config entry carries no board.
+func (landingjobs) boardless() {}
+
+// landing.jobs carries postings from many companies, so it stays in the source facet.
+func (landingjobs) aggregator() {}
+
+// landingjobsPosting is one posting, body inline (no detail call).
+//
+// The feed also carries `tags`, `type`, `gross_salary_low`/`gross_salary_high` and
+// `currency_code`, none of which are declared here. Salary has no home on Job (enrichment owns
+// it), and the element shape of `tags` and the vocabulary of `type` are undocumented and were
+// not verified live — declaring a wrong Go type for either would fail the decode of the WHOLE
+// feed, not just that field, so both are left to the pipeline's own dictionaries. Adding them is
+// a small change once the shapes are confirmed against the live endpoint.
+type landingjobsPosting struct {
+	Title            string             `json:"title"`
+	URL              string             `json:"url"`
+	Locations        []landingjobsPlace `json:"locations"`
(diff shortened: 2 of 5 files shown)


strelov1:
Live-verified against `landing.jobs/api/v1/jobs` (one-off GET, 50 postings, 2026-08-13):

**1. Multi-location postings lose data (correctness).** 14/50 (28%) sampled postings carry more than one entry in `locations[]` — e.g. one posting listed in Munich, Lisbon, AND Cologne simultaneously. Both `landingjobsLocation` and `landingjobsCountries` (landingjobs.go:169, :186) read only `p.Locations[0]`, so `Location` and the structured `Countries` reflect just the first entry. A candidate filtering by any of the other listed countries won't see this posting, even though the source states them in a structured field the adapter already parses. Worth taking all entries, not just the first — `deel.go`'s `for _, jl := range p.Job.JobLocations` is the closest existing pattern for the free-text side; `Countries` would need a small union helper since no current adapter aggregates it across multiple locations.

**2. The identity design's premise doesn't match the live payload.** The doc comment states "Neither the company nor a posting id is in the payload" — but every sampled item carries a unique numeric `id` field (verified 50/50 unique). Company genuinely isn't in the payload, so the URL is still needed for that — but building `ExternalID` from the two URL slugs instead of the numeric id is riskier than it needs to be: the sample already shows a slug with a year baked in (`…-in-lisbon-2025`), and if landing.jobs ever regenerates a slug, the current scheme would silently duplicate the posting rather than update it. Recommend keying off `id` (with company still derived from the URL for the `Company` field) rather than the slug pair.

**3. `type` (employment type) is decodable and already has a home.** Live payload's `type` is a plain string (`"Full-time"` in every sampled item), so declaring it as `Type string` carries none of the whole-feed-decode risk the PR cites for `tags`. It also maps cleanly through the existing `schemaEmploymentType` helper (schema.go:17) via the same `strings.ReplaceAll(t, "-", "_")` pattern `remotli.go:107` — the PR's own cited precedent — already uses for the identical case-format. Recommend wiring `EmploymentType: schemaEmploymentType(strings.ReplaceAll(p.Type, "-", "_"))`.

Everything else checked out: no `company` field in the payload (URL-derivation is correct), pagination is confirmed dead (page 1 == page 2, verified live), and `locations: null` for remote postings is handled without panics.

timon0305:
> Live-verified against `landing.jobs/api/v1/jobs` (one-off GET, 50 postings, 2026-08-13):
> 
> **1. Multi-location postings lose data (correctness).** 14/50 (28%) sampled postings carry more than one entry in `locations[]` — e.g. one posting listed in Munich, Lisbon, AND Cologne simultaneously. Both `landingjobsLocation` and `landingjobsCountries` (landingjobs.go:169, :186) read only `p.Locations[0]`, so `Location` and the structured `Countries` reflect just the first entry. A candidate filtering by any of the other listed countries won't see this posting, even though the source states them in a structured field the adapter already parses. Worth taking all entries, not just the first — `deel.go`'s `for _, jl := range p.Job.JobLocations` is the closest existing pattern for the free-text side; `Countries` would need a small union helper since no current adapter aggregates it across multiple locations.
> 
> **2. The identity design's premise doesn't match the live payload.** The doc comment states "Neither the company nor a posting id is in the payload" — but every sampled item carries a unique numeric `id` field (verified 50/50 unique). Company genuinely isn't in the payload, so the URL is still needed for that — but building `ExternalID` from the two URL slugs instead of the numeric id is riskier than it needs to be: the sample already shows a slug with a year baked in (`…-in-lisbon-2025`), and if landing.jobs ever regenerates a slug, the current scheme would silently duplicate the posting rather than update it. Recommend keying off `id` (with company still derived from the URL for the `Company` field) rather than the slug pair.
> 
> **3. `type` (employment type) is decodable and already has a home.** Live payload's `type` is a plain string (`"Full-time"` in every sampled item), so declaring it as `Type string` carries none of the whole-feed-decode risk the PR cites for `tags`. It also maps cleanly through the existing `schemaEmploymentType` helper (schema.go:17) via the same `strings.ReplaceAll(t, "-", "_")` pattern `remotli.go:107` — the PR's own cited precedent — already uses for the identical case-format. Recommend wiring `EmploymentType: schemaEmploymentType(strings.ReplaceAll(p.Type, "-", "_"))`.
> 
> Everything else checked out: no `company` field in the payload (URL-derivation is correct), pagination is confirmed dead (page 1 == page 2, verified live), and `locations: null` for remote postings is handled without panics.

Good catches, thanks for checking live. Fixed. 
1. Takes all locations now, not just the first. Added a small 'countriesFromCodes' helper next to 'countryFromCode' to union and dedupe the codes. 
2. ExternalID is the numeric 'id' now. You're right the slug pair was fragile. I'd built from the field list in the issue rather than the real payload, and 'id' wasn't in it. Company still comes from the URL. 
3.  EmploymentType wired as suggested. 
Left 'tags' out, still not sure of the element shape and a wrong type breaks the whole decode. Easy to add later.

PR #1887, opened 2026-08-13
Resolve real company names for boards left with a numeric platform id

strelov1:
## Summary
- SEO audit flagged that some job pages show the company as a raw internal id (e.g. `175014`) instead of a real name — visible in the page `<title>`, JSON-LD `hiringOrganization.name`, and the UI.
- Root cause: onboarding join.com boards left `company` as the bare numeric join.com company id whenever the harvest never resolved it — 2918 rows in `sources/join.yml` (plus one `bamboohr` straggler at board `415`).
- `companyname.SlugLike` now also flags a bare numeric id (it previously required a letter, so these placeholders were invisible to the existing auto-resolve backfill). `Accept` trusts a resolved name for a numeric id without demanding text overlap with the id — there is none by construction — trusting instead the new `join` resolver's authoritative lookup by that exact id against join.com's own public company-profile API (`GET /api/public/companies/{id}` → `{"name": ...}`).
- Data fix: 2913/2918 join.com boards resolved live and rewritten in `sources/join.yml`; 5 ids 404/410 on join.com's own API (company gone from the platform — left as-is, flagged for a follow-up decision on whether to prune those boards). `bamboohr` board `415` resolved to "415 Group" from its own job description.
- Left un-fixed, flagged for review rather than guessed: `breezy` board `6037` and `bamboohr` board `350` (both dead, 0 jobs), `personio` board `540` (a demo/test tenant, not a real company), `greenhouse` board `1910genetics` (no live confirmation of the full name), `teamtailor` board `500stockholm...` (platform's own registered name is genuinely "500" — may not be a bug at all).

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/companyname/... ./cmd/backfill-company-names/...`
- [x] `go run ./cmd/validate-sources sources/` — 203 files OK
- [x] `python3 -c "import yaml; yaml.safe_load(open('sources/join.yml'))"` — still valid YAML after the bulk rewrite
- [x] Spot-checked the original bug report's vacancy (join company id `175014`) now resolves to `Goodweek`

## Summary by CodeRabbit

- **New Features**
  - Improved company name resolution for Join listings using platform profile information.
  - Numeric platform identifiers can now be matched to verified company names.

- **Bug Fixes**
  - Improved ATS board detection for numeric placeholder names.
  - Updated BambooHR’s company display name to “415 Group” while preserving its board identifier.
  - Prevented resolved company names from being incorrectly treated as board identifiers.

### DIFF
--- a/cmd/backfill-company-names/main.go
+++ b/cmd/backfill-company-names/main.go
@@ -148,7 +148,7 @@ func resolveNames(ctx context.Context, rows []db.ListSlugLikeCompaniesForBackfil
 				log.Printf("resolve %s (%s): %v", row.Slug, row.Source, err)
 				return nil // a single board failing must not abort the run
 			}
-			name, ok := companyname.Accept(row.Name, candidate)
+			name, ok := companyname.Accept(row.Source, row.Name, candidate)
 			mu.Lock()
 			defer mu.Unlock()
 			if !ok {
--- a/cmd/backfill-company-names/main.go
+++ b/cmd/backfill-company-names/main.go
@@ -136,7 +136,7 @@ func resolveNames(ctx context.Context, rows []db.ListSlugLikeCompaniesForBackfil
 			stats.noSource++
 			continue
 		}
-		board, ok := companyname.BoardFromURL(row.Source, row.URL)
+		board, ok := companyname.Board(row.Source, row.Name, row.URL)
 		if !ok {
 			stats.noSource++
 			continue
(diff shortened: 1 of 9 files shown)


strelov1:
## Fixes Applied Successfully

Fixed 3 file(s) based on 1 CodeRabbit feedback item.

**Files modified:**
- `internal/companyname/gate.go`
- `internal/companyname/gate_test.go`
- `cmd/backfill-company-names/main.go`

**Commit:** `7dae8bf9`

The latest autofix changes are on the `fix-join-company-id-as-name` branch.

Issue #1761, opened 2026-08-11
harvest: echojobs — onboard orphan companies' own ATS boards


strelov1:
Snapshot (prod, 2026-08-11): **76,258 active jobs / 6,996 active companies** via `echojobs`.

Task:

1. Run `go run ./cmd/harvest-orphans -aggregators=echojobs` to list companies the catalogue holds only through `echojobs` (needs `DATABASE_URL`).
2. Live-validate candidate boards with `go run ./cmd/harvest-boards <provider> orphans.seed.json` and commit the ones that check out to `sources/<provider>.yml`.
3. Re-ingest, confirm the onboarded companies now carry their own first-party postings.
4. Decide `echojobs`'s fate: does what's left carry postings with no ATS equivalent anywhere — genuinely exclusive to this platform? Keep crawling it, narrowed to that slice. Otherwise, propose deprecating `echojobs` as a source (drop it from `sources/`, retire its cron).

Part of #1758.

strelov1:
Done, in two waves:

**Wave 1 — name-guessing (#1884, merged+deployed):** 138 new boards across 14 providers (smartrecruiters +47, pinpoint +27, jazzhr +19, bamboohr +11, greenhouse +12, ashby +6, gem +6, lever +6, trakstar +9, jobvite +4, freshteam +2, breezy +1, careerplug +1, recruitee +1). 781 first-party postings ingested.

**Wave 2 — domain-following (#1888, merged+deployed):** of the ~2045 orphans left, 329 had a known homepage on file. Ran `cmd/harvest-ats resolve` against those instead of guessing slugs — reaches tenant-keyed platforms (workday, oracle) name-guessing structurally can't. 30 more boards: workable +20, workday +3, oracle +5, greenhouse +1, jazzhr +1. 784 first-party postings ingested.

**Step 4 (echojobs's fate):** not resolved — most of the ~2000 remaining orphans are "not yet found", not "confirmed exclusive": `join`/`personio`/`workable` hit rate limits mid-run and wrote nothing (retry with `-pace`), the domain-following pass only had a homepage on file for 329/2045, and slug-guessing only reaches ~4% by design. Filed as a follow-up rather than guessing at echojobs's fate from an incomplete picture.

Bugs found and fixed along the way:
[...]

PR #1891, opened 2026-08-13
harvest-boards: fold case in SmartRecruiters dedup, reject non-numeric apploi slugs

strelov1:
## Summary
Two dev-tool bugs surfaced live while onboarding echojobs orphan companies (#1884, #1888):

- **SmartRecruiters dedup was case-sensitive.** Greenhouse/Ashby/Workday already fold board-id case before comparing against existing entries; SmartRecruiters didn't, so a name-derived candidate differing only in case from an already-onboarded board (`archirodongroup` vs the stored `ArchirodonGroup`) read as new and got appended. Happened twice — 14 duplicates in #1884, 6 more in #1888 — caught both times by CI's `validate-sources` and dropped by hand before merge. Now `smartRecruitersProber` implements `dedupKey` like the other three.
- **apploi's prober trusted a slug it shouldn't have.** Its own doc comment says board = numeric employer id, but nothing enforced that — fed a name-derived candidate, `api.apploi.com/v1/jobs?employer=<slug>` silently ignores a value it can't parse and answers a non-empty "live" jobs list regardless. A 5,216-candidate name-derived seed read as a **100% hit rate**, confirmed live and reverted before commit in #1884 (never reached prod). Now rejects any non-numeric slug before making the request.

## Test plan
- [x] `go test ./cmd/harvest-boards/...`
- [x] New tests: `TestDedupKeyOf` extended for smartrecruiters; `TestApploiProberRejectsNonNumericSlug` / `TestApploiProberAcceptsNumericSlug` added — the reject test uses a canned response that WOULD read as a live board if the guard regressed, so it fails loudly rather than passing vacuously.
- [x] `go vet ./...`, `gofmt -l .` clean.

Follow-up from #1761/#1758.

### DIFF
--- a/cmd/harvest-boards/boardfile_test.go
+++ b/cmd/harvest-boards/boardfile_test.go
@@ -50,6 +50,12 @@ func TestDedupKeyOf(t *testing.T) {
 	if dedupKeyOf(ashbyProber{})("Ramp") != "ramp" {
 		t.Error("ashby key must fold case")
 	}
+	// SmartRecruiters' postings API is case-insensitive too (confirmed live twice, #1884/#1888:
+	// a name-derived candidate slipped past dedup because it only differed in case from an
+	// existing board), so the dedup key folds case like the others.
+	if dedupKeyOf(smartRecruitersProber{})("ArchirodonGroup") != "archirodongroup" {
+		t.Error("smartrecruiters key must fold case")
+	}
 }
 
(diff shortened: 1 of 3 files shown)


strelov1:
Superseded — this fix landed already via #1895 (which cherry-picked the same commit before adding the URL-extraction harvest results on top). Nothing left to merge here.

Issue #1873, opened 2026-08-13
Investigate OCF (Open Career Format) export/import onboarding


strelov1:
## Context

[OCF (Open Career Format)](https://opencareerformat.org) models "the career behind the résumé" — a candidate-owned career-memory schema (experience, projects, skills, competencies, goals, provenance per item), distinct from a rendered CV. Its [career-ops integration mapping](https://opencareerformat.org/mappings/career-ops-integration.html) describes a **standalone export skill** pattern: generate downstream user-layer files (`cv.md`, `config/profile.yml`, `modes/_profile.md`, `interview-prep/story-bank.md`) from an OCF master, with no plugin registry approval needed and no operational state imported back.

This matters for us because `santifer/career-ops` has twice declined listings-side integration with freehire on policy grounds (aggregation is reserved first-party — see #1082/#1197, #2350). The *format* side was explicitly left open in that thread (discussion #1342): publishing/consuming OCF needs nobody's permission, unlike proposing a plugin that surfaces vacancies.

## What to think through

- **Export**: map `internal/experience` (durable employments + evidence atoms) and `resume_structured` onto OCF's schema (v0.3, `https://opencareerformat.org/v0.3/schema.json`). Key constraint: OCF's `sourceArtifacts`/provenance fields are advisory only ("the schema defines them but does not enforce how tools honor them") — our provenance enum (`cv_import` / `stated_in_chat` / `manual` vs `agent_inferred`) must not get flattened on export; only candidate-asserted facts should read as OCF-trustworthy.
- **Import**: whether/how an OCF document from another tool could seed a new user's experience bank, and how that reconciles with our own provenance gate (agent-inferred facts still can't be written into a CV).
- **Onboarding path**: standalone one-way export skill first (matches the career-ops mapping precisely, zero coordination cost) vs. a versioned bidirectional `*-plugin-ocf` connector later.
- **Versioning risk**: OCF is v0.3, single maintainer, near-zero adoption today — pin the schema version and treat this as speculative until adoption evidence shows up.

## Non-goals

- No listings/aggregation integration with career-ops — that door is explicitly closed (core-or-plugin-registry-alike).
- No import of career-ops *operational* state (job scores, pipeline status) — only reviewed career-memory content flows back, per the mapping doc.

This is a research/design ticket, not an implementation one — the output should be a decision on whether OCF export is worth building, not code.

strelov1:
[...]

PR #1904, opened 2026-08-14
docs(design): OCF export/import research pass

timon0305:
Research pass for #1873, which asks for a decision rather than code.

**Decision: not yet — note the seam, don't build.** Two named triggers to reopen, below.

## The adoption hunch checks out, and it's worse than "near-zero"

`opencareerformat/opencareerformat`, checked 2026-08-14:

| Signal | Value |
|---|---|
| Stars | 5 |
| **Forks** | **0** |
| Watchers | 2 |
| Created | 2026-05-22 (under three months old) |
| Last push | 2026-08-04 |

Stars are a bookmark. **Zero forks is the number that matters** — nobody has copied the schema, so nobody has implemented against it. No published list of adopting tools, and no stated stability policy for a v0.3.

The schema itself is real and decent, for the record: it models the same thing `internal/experience` does (career memory, not a rendered CV), nests positions under one tenure, and carries skill taxonomies (O\*NET/ESCO/SFIA). The conceptual fit is genuinely good, which is why this got a written answer rather than a one-line decline.

## The finding worth keeping either way

**The two provenance models are philosophically incompatible, and the near-miss is the hazard.**

freehire's is a binary, enforced, fail-closed gate — `Publishable()` decides whether a claim may reach a CV, and an unknown value fails closed. OCF's is advisory annotation spread across four fields (`meta.source.kind`, an open-shape per-item `provenance`, `reviewStatus`, `visibility`), and per the ticket's own quote the schema "defines them but does not enforce how tools honor them."

`reviewStatus` looks like the counterpart and isn't: it records *whether a human checked an item*, where our enum records *who originated it*. An unconfirmed `agent_inferred` atom is fairly called `unreviewed` — but so is a `cv_import` atom, which nobody reviewed and which is fully publishable. Mapping the gate onto it loses the distinction **in both directions at once**.

**Resolution: export by omission, not annotation.** Publishable atoms only; leave `agent_inferred` material out of the document entirely rather than shipping it with a warning a consumer is free to ignore — which would relocate the exact harm the gate prevents to a tool with no gate at all.

That's already the house style. `ghost.ContributorGate` withholds a count rather than serving it redacted, and the reasoning at that constant applies verbatim: absence is what makes the guarantee structural, because no later caller can forget to redact what was never served.

## Also in the doc

- **Mapping table** — small and mechanical, thanks to the boundary table already in `internal/experience/AGENTS.md`. Two fields have no counterpart at all: `sourceArtifacts` (our provenance is an enum, not a pointer to a retained artifact) and `competencies` (we derive nothing like it).
- **`goals` maps to `internal/userprofile`** — not mentioned in the ticket, and it's where the operational-state non-goal bites: an import would have to deliberately drop the targeting half.
- **Import is materially harder and shouldn't ship with export.** Every inbound atom has an origination our enum can't express, so it needs a fifth non-publishable value plus a confirmation flow. The inbound `reviewStatus` must be refused, for the same reason the export can't rely on it — trusting it inbound while distrusting it outbound would be incoherent.
- **Why cheapness alone doesn't justify building it**, given the mapping is about a day: AGENTS.md's "note the seam for later instead"; `/me/resume` and `/me/profile` already meet the portability need (this is one third-party *encoding* of portability we have, not new capability); and the motivation points largely at feeding career-ops, which has twice declined to integrate.

## Triggers to reopen

- **A second independent implementation appears** — a fork, or any OCF reader that isn't the spec's own author.
- **A user asks.** One real request beats format-watching, and the mapping means the answer is "a day", not "a project".

Docs-only. Every referenced path and quoted symbol (`Publishable()`, `ContributorGate`, `Professional`) verified against the tree.

### DIFF
--- a/docs/superpowers/specs/2026-08-14-ocf-export-design.md
+++ b/docs/superpowers/specs/2026-08-14-ocf-export-design.md
@@ -1,0 +1,180 @@
+# OCF (Open Career Format) export/import — research pass
+
+Research pass for [#1873](https://github.com/strelov1/freehire/issues/1873). The ticket asks
+for a decision on whether OCF export is worth building, not code.
+
+**Decision: not yet. Note the seam, write down the mapping, and don't build.** Two named
+triggers below are what should reopen it. The interesting part of this pass is not the mapping —
+that is small — but that OCF's provenance model and freehire's are philosophically incompatible,
+and the resolution turns out to be cheap and worth recording before anyone starts.
+
+## What OCF is, verified
+
+`opencareerformat.org`, v0.3, last site update 2026-08-04. The schema is real and reasonably
+thought through: a `person` + `meta` root, then `skills`, `competencies`, `experience`,
+`projects`, `education`, `certifications`, `languages`, `goals`, `sourceArtifacts`, and a long
+tail (`governance`, `teaching`, `speaking`, `patents`, `publications`, `awards`, …). Experience
+nests positions under one organization/period so a promotion inside one employer stays one
+tenure. Skills carry a category, proficiency, date range, aliases and taxonomy URIs (O*NET /
+ESCO / SFIA).
+
+It models the same thing `internal/experience` does — the career behind the CV rather than a
+rendered CV — so the conceptual fit is genuinely good. That is what makes this worth a written
+answer rather than a one-line decline.
+
+## Adoption, measured
+
+The ticket flags "near-zero adoption today" as a hunch. It checks out, and the numbers are worse
+than the word "near-zero" implies:
+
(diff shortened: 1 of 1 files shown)


strelov1:
Спасибо за тщательное исследование — особенно за разбор несовместимости моделей провенанса, это ценная находка независимо от решения по OCF. Закрываем PR и issue #1873 без мёржа: решение задокументировано ("not yet"), но конкретной потребности строить сейчас нет. Триггеры на переоткрытие остаются в силе (появление второй независимой реализации OCF или реальный запрос пользователя).

PR #1905, opened 2026-08-14
Replace Meilisearch semantic search with pgvector; drop /recommendations

strelov1:
## Summary

Removes the Meilisearch-backed `jobs_semantic` hybrid-search index entirely and replaces its two real consumers with a Postgres/pgvector-based approach:

- **`/jobs/:slug/similar`** — now a precomputed lookup (`jobs.similar_job_ids`, filled offline by a new `cmd/similar-backfill` worker) instead of a live Meili query. Excludes same-company matches. No live vector math on the request path anymore.
- **`/me/recommendations` (CV-based)** — **removed outright**, not migrated. Investigating the pgvector migration surfaced real complexity (no SQL facet-filter translator exists anywhere in this codebase; building one would duplicate ~20+ facets' worth of already-trusted Meili filter logic) for a feature of uncertain usage, so it was dropped instead. The "Recommended" feed sort mode and the CV-embedding write path on résumé upload go with it.
- **Job descriptions are now embedded as HTML-stripped, full-length text split into chunks** (`job_semantic_chunks`, one row per chunk) instead of a single truncated, HTML-laden vector — a real quality fix folded in along the way (only ~15-20% of an average description was ever reaching the embedder before).

Full design rationale, including two mid-implementation reversals (adopting chunked embeddings, dropping recommendations instead of migrating it) and the real bugs caught during review (an N+1 delete query, a missing partial index, a `company_slug` NULL-handling bug in the nearest-neighbour query, a wasteful extra TEI call per job, several pieces of orphaned dead code): `openspec/changes/drop-hybrid-search-pgvector-similar/{proposal,design}.md`.

**Not included — deliberately deferred to manual, step-by-step production operations** (each requiring explicit confirmation, not part of this PR): installing the `pgvector` Postgres extension on prod, applying the migration, triggering the full-catalogue re-embed (a real multi-hour-to-multi-day TEI cost), building the HNSW index, deploying `cmd/similar-backfill`, and finally dropping the live `jobs_semantic` Meili index. See `tasks.md` sections 3 and 8.

## Test plan

- [x] `go build ./...` / `go vet ./...` / `go vet -tags=integration ./...` clean
- [x] `go test ./...` clean
- [x] `go test -tags=integration ./...` (full tagged suite) clean
- [x] `pnpm --dir web run check` clean (0 errors)
- [x] `pnpm --dir web test` clean (940 tests)
- [x] Every task reviewed for spec compliance and code quality before being marked done (see commit history / `tasks.md`)

## Summary by CodeRabbit

* **New Features**
  * Added precomputed similar-job results, excluding same-company jobs and unavailable listings.
  * Added chunked job-description embeddings to improve similarity matching.
  * Added an endpoint to reset a base CV from résumé data.
  * Added `software_engineering` and `preparing` to supported API values.

* **Changes**
  * Search is now keyword- and facet-based; semantic controls were removed.
  * Removed CV-based recommendations and Recommended feed sorting.

* **Documentation**
  * Updated API and architecture documentation for revised search and similarity features.


strelov1:
## Fixes Applied Successfully

Reviewed all 13 CodeRabbit feedback items and applied fixes for 10 of them; the remaining 3 were skipped as deliberate, already-documented design deferrals (the ANN index build and the candidate-side freshness policy are explicitly scheduled as manual prod operations in design.md, not gaps).

**Files modified:**
- `internal/search/client.go` — kept `embedderModel` on its pre-chunking identity so this deploy doesn't trigger the full-catalogue re-embed early; that bump is now its own separate, scheduled step
- `internal/db/queries/jobs.sql` — invalidate a job's precomputed similar-jobs list whenever its `company_slug` actually changes (UpsertJob, UpdateJobCompany, UpsertManualJob, UpdateManualJob)
- `internal/db/queries/semantic.sql` — guard the similar-jobs backfill's write against a race with a concurrent re-embed (a chunk-generation check, conditional update)
- `internal/similarjobs/runner.go`, `cmd/similar-backfill/{main.go,store.go}` — thread the generation guard through, validate CLI flags, track the new stale-write outcome distinctly from success/failure
- `internal/search/chunk.go` — split a single oversized token (e.g. a long URL) instead of letting it overrun the embedding chunk budget
- `internal/search/embed.go` — cap chunk count per job to stay within `int16`
- `cmd/embed/embed_integration_test.go` — extended the re-embed test to also cover a shrinking chunk set
- `docs/API.md`, `openspec/changes/drop-hybrid-search-pgvector-similar/{proposal.md,specs/semantic-embedding/spec.md}`, `cmd/embed/main.go` — corrected stale doc/comment claims left over from the legacy embedding-column removal and the dropped `/recommendations` endpoint

**Commit:** `9b502a8f`

The latest autofix changes are on the `worktree-hybrid-search-default-ratio` branch.

PR #1947, opened 2026-08-15
Associate the development signing team too

strelov1:
Third and last piece of the mobile OAuth return leg (after #1936 and #1942), found by running the flow in the simulator.

The association named only the EAS team, `25U9HN34VM`. A build produced by Xcode or `expo run:ios` is signed with the personal team `WYHG3T94Q4`, so the OS refused it — `swcd` matched the domain, found no entry for that app id, and the auth session died with:

```
Application with identifier me.freehire.mobile is not associated with domain freehire.me
```

Both teams own the same bundle id and both are ours, so the file now lists both. Exact app ids, no wildcard — a local development build can finish the OAuth return, and nothing else gains access.

Together with the two earlier PRs the return leg is complete: the path is claimed, `webcredentials` is present, nginx serves the file as `application/json` (that fix lives in freehire-ops), and both signing identities are recognised.

### DIFF
--- a/web/static/.well-known/apple-app-site-association
+++ b/web/static/.well-known/apple-app-site-association
@@ -3,7 +3,7 @@
     "apps": [],
     "details": [
       {
-        "appIDs": ["25U9HN34VM.me.freehire.mobile"],
+        "appIDs": ["25U9HN34VM.me.freehire.mobile", "WYHG3T94Q4.me.freehire.mobile"],
         "components": [
           { "/": "/jobs/*" },
           { "/": "/auth/mobile-callback*" }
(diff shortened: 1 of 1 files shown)


strelov1:
Not needed, and it would have been the wrong fix.

Both Apple Development certificates in the keychain belong to 25U9HN34VM — the team that owns the App ID, the Sign in with Apple key and the Expo project. WYHG3T94Q4 was never ours to begin with: Expo picked it as the default DEVELOPMENT_TEAM during prebuild.

The right fix is to pin the team in the app config (`ios.appleTeamId`), so local builds are signed by the same team the association already names. Widening a production association file to accommodate a stray signing identity would have been backwards.

PR #1985, opened 2026-08-16
Recognize UKG and CatsOne boards in the shape their adapters crawl

strelov1:
## The bug

`atsboard` maps a job URL to `(source, board)`, and the board id it returns must be the one the
ingest adapter keys on. For two platforms it was not, so **every board discovered from a URL was
unusable** — recognized, recorded, then rejected by the crawl.

**UKG.** The adapter takes `<host>/<tenant>/<guid>` and refuses anything else
(`internal/sources/ukg.go`: `board %q must be <host>/<tenant>/<guid>`). The table used `path`
mode, which returns the first path segment — the tenant code alone, e.g. `ADV1023AVR`. It also
named a single host, `recruiting.ultipro.com`, where the catalogue actually holds:

| host family | boards |
|---|---:|
| `*.ultipro.com` | 2166 |
| `*.rec.pro.ukg.net` | 709 |
| `*.ultipro.ca` | 164 |

The other three were not recognized at all.

**CatsOne.** The adapter fetches `https://<board>/careers`, so the board is the whole host, which
is what `catsone.yml` stores (`authoritypartnersinc.catsone.com`). The table used `subdomain`
mode and returned the bare label.

## The fix

A `hosttenantboard` mode for UKG that reads `<host>/<tenant>/JobBoard/<guid>`. The literal
`JobBoard` segment is required: it is the only thing distinguishing a board URL from the rest of
a tenant's site, and without it there is no guid to take, so such a URL is declined rather than
turned into an id the crawl will reject.

The id is lower-cased. UKG answers either spelling — verified live against a real tenant, upper
and lower, both 200 — and every board in the catalogue is stored lower-case, so keeping a link's
own casing would file a board we already crawl as a new one. (This is the opposite of Workday
just above it, whose site segment is a human-chosen name the ingest stores verbatim — hence the
comment on each.)

CatsOne moves to `host` mode. Its custom-domain tenants (`jobs.evoplay.com.ua`) stay underivable
from a URL, like every other custom-domain ATS in this table.

## How it surfaced

The August 2026 website-walk harvest (#1963, #1982) proposed 87 UKG candidates and a handful of
CatsOne ones. Every one probed dead — not because the boards were dead, but because the ids were
never boards. Three platforms share the pattern and it is worth stating plainly: **a detector and
an adapter that disagree about the id shape fail silently**, as a harvest that finds nothing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved job board recognition for UKG-hosted career sites across US, Canadian, and regional domains.
  * Added support for extracting UKG board identifiers from host and tenant paths, with consistent lowercase formatting.
  * Improved CatsOne board recognition using the complete host address.
  * Removed outdated recognition for unsupported UKG and CatsOne URL formats.

### DIFF
--- a/internal/atsboard/board.go
+++ b/internal/atsboard/board.go
@@ -56,6 +56,12 @@ const (
 	// numeric institution id, so its localisation and section segments (/cw/en/search) are not
 	// boards — reading one as a board names an institution that does not exist.
 	modePathNumeric = "pathnumeric"
+	// hosttenantboard: board = "<host>/<tenant>/<guid>" from <host>/<tenant>/JobBoard/<guid>/….
+	// UKG addresses a board by all three: the host carries the data-residency TLD, the tenant is
+	// its customer code and the guid picks one of the tenant's boards. The adapter needs the same
+	// three (internal/sources/ukg.go), so anything less is not a shorter board id — it is one the
+	// crawl rejects outright.
+	modeHostTenantBoard = "hosttenantboard"
 )
 
 // atsBoards lists the supported multi-tenant ATS: a host (exact or subdomain-suffix match) →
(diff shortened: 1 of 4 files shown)


strelov1:
Pushed the harvest this unlocks: **75 boards** (ukg 58, catsone 17), from re-walking the 257 companies whose careers pages had produced an unusable id.

The run also confirms the shape end to end: of the 206 UKG ids proposed, **140 matched boards already in `ukg.yml`** and were dropped as existing. Before the fix not one matched — the ids were simply a different id space.

No duplicate board id, exact or case-variant, in either touched file.

PR #1990, opened 2026-08-16
Page both sitemaps from the search index, not a row_number() walk

strelov1:
## Why

`/sitemap.xml` returns **500 on prod**. It is only reachable at all because nginx keeps serving a stale cached copy (`x-cache-status: UPDATING`); a cache-busted request answers 500 in 10.7s.

Building the sitemap index needs the boundary between chunks, and both boundary queries numbered every eligible row with `row_number()` — a walk that cannot stop early. Measured on prod 2026-08-16:

```
job boundaries:   64s over 3.4M rows   (SSR fetch timeout is 10s)
Heap Fetches:     762,341              ("index-only" in name only — autovacuum trails the ingest)
```

The migration that introduced the job query sized it for ~1.1M rows. The open, non-duplicate set is now 3.4M.

## What changed

Both sitemaps page a **Meilisearch index by offset** instead. The engine reports its document count on every response, so chunk boundaries become arithmetic — no query at all — and it addresses any offset directly:

```
offset 0        → 0.06s
offset 1.2M     → 0.19s
full 25k page   → ~2.5s
```

The index also holds a **better set** than the table did — open, non-duplicate, non-private, categorized. The table's equivalent scope is 2.7x larger and includes 2.1M postings the site's own search cannot find, plus 10 private ones that were leaking into the sitemap.

`CompanyDocument` gains `updated_at`, solely to emit `<lastmod>`. It is deliberately not searchable, filterable, or sortable.

## Deploy order

Safe in either order, but the sitemap is better after both:

1. **Deploy this.** The job sitemap is correct immediately. Company URLs ship **without `<lastmod>`** until step 2 — the read path treats a missing lastmod as normal, so nothing breaks.
2. **Run `reindex-companies`** to backfill `updated_at` into the companies index. Then company `<lastmod>` appears.

`jobs_sitemap_idx` (migration 0107) is now orphaned in code — nothing queries it. Left in place on purpose: dropping a partial index over 3.4M rows is an ops call, and it may still serve other reads with the same predicate. Worth a separate look.

## Verification

Against the **live prod indexes** over a read-only tunnel:

| | jobs | companies |
|---|---|---|
| documents | 1,263,112 | 292,271 |
| chunks | 51 | 30 |
| tail page | 13,112 docs | 2,271 docs |
| offset past end | empty, 209ms | empty, 212ms |

Also: `go test ./...`, `go vet -tags=integration ./...`, `gofmt`, `golangci-lint` (clean on the changed files), eslint on the changed web files.

## Tests

The job and company sitemap tests **no longer need Postgres** — they run against a stub index. What still needed a real router (that `/jobs/sitemap` is not swallowed by `/jobs/:slug`) is now a unit test asserting through the handler's own `register()`, so it pins the real route order. Both integration test files are deleted; the SQL they covered is gone.

New coverage worth naming:

- **`omitzero`, not `omitempty`** on the lastmod. `omitempty` does nothing to a `time.Time` (a struct is never "empty"), so a company without one would have shipped `<lastmod>0001-01-01T00:00:00Z</lastmod>` to every crawler. There is a test on exactly that.
- Each half reads **its own** index — nothing else in their shared shape would catch them being wired to the same one.
- A stale or junk offset (including past `int32`) serves an empty page, never an error.

## Summary by CodeRabbit

* **New Features**
  * Sitemap pages now load from search indexes for improved offset-based pagination.
  * Job and company sitemap URLs now use numeric `offset` parameters.
  * Sitemap entries can omit modification dates when unavailable.
  * Missing search indexes return a service-unavailable response.

* **Bug Fixes**
  * Invalid, negative, or stale offsets are handled safely without errors.
  * Sitemap routes are protected from conflicts with other URL routes.


strelov1:
## CodeRabbit Autofix Review Complete

Reviewed 2 CodeRabbit feedback items, both applied in `dd703a43`.

**Files modified:**
- `internal/handler/sitemap.go` — floored `?chunk=` at 1000
- `internal/handler/sitemap_test.go` — boundary tests moved to multiples of the floor; new test that the floor clamps rather than truncates
- `internal/search/sitemap.go` — corrected the SitemapDocument projection comment
- `web/src/lib/api.ts` — corrected the Sitemap section comment

Validated: `go test ./...`, `go vet -tags=integration ./...`, `golangci-lint` (clean on changed files), eslint.

PR #2016, opened 2026-08-16
Stop losing and showing job descriptions that never arrived

### LINKED ISSUE #1866: Find and investigate jobs with empty descriptions
Some jobs in the catalog may end up with an empty (or near-empty) `description` field — likely from source parsers that fail to extract the body, or enrichment steps that clear it. Need to:

- Query `jobs` for empty/whitespace-only descriptions and get a count/breakdown by `source`
- Determine whether this is a parser bug (per-source) or a downstream issue (dedup/enrich)
- Decide handling: backfill, re-ingest, or soft-close affected postings

strelov1:
Closes #1866.

## What the catalogue actually holds

15,816 live jobs carry an empty `description` — 0.48% of the 3,305,695 open, non-duplicate rows. Measured on prod, 2026-08-16.

| source | empty | live | note |
|---|---:|---:|---|
| oracle | 4,883 | 205,996 | source publishes no body |
| paylocity | 3,433 | 23,358 | rows no longer re-crawled (see below) |
| hh | 2,538 | 9,382 | hydrate-once |
| smartrecruiters | 1,066 | 220,270 | adapter dropped the only text |
| paycom | 900 | 69,679 | source publishes no body |
| jobdanmark | 526 | 7,404 | not re-crawled |
| justjoin | 368 | 9,036 | hydrate-once |
| clinch | 356 | 363 | by design (WAF latch) |
| personio | 340 | 35,343 | |
| workday | 277 | 747,370 | |

Not a regression: new empties arrive every week across every top source, and the oldest go back to the catalogue's own history.

## Three causes, one non-cause

**1. smartrecruiters dropped the only text there was.** The adapter excluded `companyDescription` as boilerplate. Probing the API for one of these postings shows `jobDescription`, `qualifications` and `additionalInformation` all empty and `companyDescription` carrying the entire ad. The exclusion is now a *preference*: the company section is used when the role sections are empty.

**2. Hydrate-once made a single failure permanent.** The hydrating sources fetch detail only for postings the catalogue lacks, and being stored is what marks a posting seen. A posting whose one detail fetch failed was therefore never retried — its body was missing for the row's whole life. A row with no description is now withheld from the seen-set until it is `pipeline.HydrationRetryWindow` (14 days) old.

The window is bounded on purpose: the *other* reason a body is missing is that the source published none, and retrying those forever would spend a detail request per crawl on each one. hh alone would be 2,538 extra requests every crawl, permanently.

**3. Not our bug — but not something to show either.** Probing Oracle's detail API live for three of the empty rows returns every description field (`ExternalDescriptionStr`, `ExternalResponsibilitiesStr`, `ExternalQualificationsStr`, `CorporateDescriptionStr`, `ShortDescriptionStr`) at length 0. The employer published a requisition with no body. Those rows stay in the catalogue — they are real postings and the lifecycle owns their closure — but they leave the search index, alongside the category-unresolved rows already excluded there. A vacancy page with a title and nothing under it is not a listing anyone can act on.

`search.DescriptionMissing` tests the **visible** text (`stripToPlainText`), not the raw column: a source with an empty rich-text field serves `<p>&nbsp;</p>`, which the ingest sanitizer legitimately keeps. It is wired into the same three places `CategoryUnresolved` already was — `cmd/reindex`'s `splitJobs`, `cmd/search-drain`'s `IndexBatch`, `internal/linkimport`. Self-healing: a row re-enters the index the moment a crawl fills its body, with no backfill step.

**Non-cause checked and cleared:** the paylocity parser. Fetching a real detail page for one of its empty rows and running `paylocityDescription` over it returns 3,514 bytes of body, and the page fetches 200 from the prod host too. Those rows are empty for the reason in the next section, not a parse failure.

## What this does not fix

The window is measured from `created_at`, so most of the existing hydrating-source backlog (created 3–4 weeks ago) is past it and will not be re-fetched. Roughly 700–800 of the freshest rows self-repair; the rest are simply hidden. That is the decided handling, not an oversight.

## Found alongside, filed separately

`paylocity` has 17,180 of 23,358 live rows unseen for more than 3 days, `arbeitsagentur` 4,499 of 4,499, `uber` 606 of 606 — open in the catalogue but no longer re-crawled, with the 48h unseen sweep not closing them. That is a lifecycle bug, larger than this one, and out of scope here.

## Verification

```
gofmt -l .                                            clean
go build ./... && go vet ./...                        ok
go vet -tags=integration ./...                        ok
go test ./...                                         pass
go test -tags=integration ./internal/db/ \
  -run TestExistingExternalIDs                        4/4 pass (real Postgres)
golangci-lint run                                     0 issues
```

New tests: two on the smartrecruiters fallback (fires when role sections are empty, does **not** fire when they are not), two integration tests on the seen-set withholding (within the window and aged out, provider-wide and board-scoped), `TestDescriptionMissing` on the visible-text rule, and `TestSplitJobs_BodylessJobsDeletedNotIndexed` so the index gate cannot be dropped silently.


strelov1:
Second commit adds the repair path for the existing backlog.

The window is measured from `created_at`, so everything that accumulated before the retry existed has already aged past it — the fix above prevents new body-less rows but does not fill the old ones. `HYDRATION_RETRY_DAYS` widens the window for one deliberate run:

```
HYDRATION_RETRY_DAYS=365 ingest sources/hh.yml
```

That re-offers every body-less row of the provider to its adapter, at one extra detail request each. Expect to repeat it — a run that hits its systemd timeout still persists what it hydrated, and the rows it fixed leave the set, so successive runs shrink the backlog rather than redoing it.

Deliberately **not** a new worker. The pipeline already knows how to hydrate a posting; the only thing standing between it and these rows is that being stored is what marks a posting seen. Widening the window is the entire repair — a separate worker would have had to duplicate every adapter's detail fetch and parse.

An unparseable or non-positive value fails the run instead of falling back to the default: this is typed by hand for a one-off, and a quiet fallback would look exactly like a repair that found nothing to repair.

Verified: `go test ./...`, `go vet -tags=integration ./...`, `go test -tags=integration ./cmd/ingest/` (9.6s, real Postgres), `golangci-lint run ./cmd/ingest/...` → 0 issues.

PR #2043, opened 2026-08-17
SEO: a heading for the API reference, and collection links from every page

strelov1:
The remaining on-page items from the freehire.me audit. Follows #2035.

## The API reference had no heading

All 140 `/docs/api/:group/:endpoint` pages shipped without an `<h1>` — the endpoint's summary was a `<p>`. That summary is the right text for the heading: the method and path are the identifier and already carry the `<title>`, while "Full-text + faceted search over open jobs." is what the page is *about*. Only the element changes; the class list keeps the size it had.

`/docs/api` is the site's fourth-largest landing page by real organic traffic, so this is not a hypothetical.

## Collection landings had no links from the homepage

Collection landings are the programmatic-SEO surface, and their only internal links came from the `/collections` hub and the job-detail see-also block. Parsing every `href` on the homepage:

| Target | Links |
|---|---|
| `/jobs/*` | 20 |
| `/features/*` | 4 |
| `/insights/*` | 1 |
| **`/collections/*`** | **0** |

`popularCollectionLinks()` resolves the existing curated pool through `collectionBySlug` — the same resolver the landing route and the sitemap use — and the footer renders it, which reaches every page rather than only the homepage. Unresolvable slugs are dropped: a footer is the worst place to advertise a 404.

## Titles and descriptions that wasted their width

- `Trends · freehire` — 17 characters naming no subject, so nothing in it could match a query → `Tech hiring trends — jobs added and removed daily · freehire`
- `Companies · freehire` — 20 characters → `companiesPageTitle` leads with the live count, like `companyPageTitle` does for one employer, and falls back to the bare subject on a zero or absent count ("0 companies hiring in tech" is an argument against clicking)
- `/about`'s description was 215 characters and got cut mid-sentence, where Google chose rather than where we would → trimmed to ~155

## A correction to the audit

`/collections` links **all 101** collections, not "101 of 198" as first reported. 198 was the total URL count of `sitemap-pages.xml`, which also holds `/about`, `/companies`, the 79 blog posts and the rest. The hub was never missing any, and there is nothing to fix there.

## Verification

Against the production API with the dev server: one `<h1>` per endpoint page holding its summary; 10 collection anchors on `/`, `/companies` and `/docs/api`; `/companies` titled "148,973 companies hiring in tech · freehire"; `/about`'s description at 156 characters.

\#2035's guards still hold — `remote-latam` is 200 through page 474 and 404 at 475 (9,465 matches, the boundary having moved with the catalogue), `ooo-marta` still `noindex`, `stripe` still not.

Green: `npm test` (1070), `npm run check` (0 errors), oxlint + eslint (0 errors).

## Not in this PR

Core Web Vitals are now measured (local Lighthouse, mobile) and **LCP fails**: 5.3 s on the homepage, 7.8 s on a job page, against a 2.5 s threshold. CLS (0.066 / 0) and TBT (60 / 120 ms) both pass, and the SEO category is 100.

It is not a byte problem and not a blocking-resource problem. Render delay is 91–94% of LCP while every critical resource completes within 805 ms unthrottled, and the mechanism is visible in the HTML: the nonce'd bootstrap does exactly **two** dynamic imports (`start.js`, `app.js`), there are **zero** `<link rel="modulepreload">` or `preload` hints, and 91 JS requests — 71 of them under 3 KiB — are discovered only by recursively parsing those two. On a 150 ms RTT link that serializes into seconds.

I did not determine why SvelteKit emits no `modulepreload` here: there is no `preload` filter in `hooks.server.ts` and no `kit.output` config, so the default should apply. Separately, `logo.freehire.me` serves 512×512 PNGs (one is 200 KB for a thumbnail slot) and ignores every size parameter — that proxy is not in this repo.

Both deserve their own change. And the field impact is unconfirmed: these are *simulated* lab numbers, and settling what real users see needs CrUX or Search Console, neither of which is connected.

## Summary by CodeRabbit

* **New Features**
  * Added server-rendered “Popular collections” links to the footer, showing only valid collection pages.
  * API reference pages now display endpoint summaries as visible page headings.

* **SEO Improvements**
  * Improved companies page titles with formatted job counts and singular/plural wording.
  * Updated About and Trends page titles and descriptions for clearer search results.

### DIFF
--- a/openspec/specs/web-ssr-seo/spec.md
+++ b/openspec/specs/web-ssr-seo/spec.md
@@ -247,7 +247,7 @@ collection landing page SHALL be enumerated in the sitemap.
 
 ### Requirement: Public list pages carry a primary heading
 
-The public list pages (`/jobs` and `/companies`) SHALL each render exactly one visible `<h1>` describing the page's content, so the primary heading is present for search engines and assistive technology.
+The public list pages (`/jobs` and `/companies`) SHALL each render exactly one visible `<h1>` describing the page's content, so the primary heading is present for search engines and assistive technology. Every API-reference endpoint page (`/docs/api/:group/:endpoint`) SHALL likewise carry one, taken from the endpoint's summary — the method and path are the identifier and already carry the `<title>`, while the summary is what the page is about.
 
 #### Scenario: The jobs list has a top-level heading
 
@@ -302,3 +302,26 @@ company.
 - **WHEN** `GET /companies/:slug` is requested for a company whose job search
   returns results
 - **THEN** the `<head>` contains no `robots` meta tag
+
+#### Scenario: An API endpoint page has a top-level heading
+
+- **WHEN** `GET /docs/api/:group/:endpoint` is requested for a documented endpoint
+- **THEN** the returned HTML body contains exactly one `<h1>`, holding that
+  endpoint's summary
+
+### Requirement: Collection landing pages are linked from every page
+
+The site footer SHALL render the popular collections as real `<a href>` anchors to
+their `/collections/:slug` landing pages, in the server-rendered HTML. Collection
+landings are the site's programmatic-SEO surface, and their only internal links came
+from the `/collections` hub and the job-detail "see also" block — the homepage, the
+strongest page on the domain, linked twenty jobs, four feature pages and no
+collection at all. Every slug SHALL resolve through the same resolver the landing
+route and the sitemap use, and an unresolvable slug SHALL be dropped rather than
+linked: a footer is the worst place to advertise a 404.
+
+#### Scenario: Any page links the popular collections
+
+- **WHEN** any public page is requested
+- **THEN** its HTML contains `<a href>` anchors to several distinct
+  `/collections/:slug` landing pages
--- a/web/src/lib/collections.test.ts
+++ b/web/src/lib/collections.test.ts
@@ -5,6 +5,7 @@ import {
   collectionBySlug,
   collectionSlugs,
   jobFacetsFromJob,
+  popularCollectionLinks,
   relatedCollectionLinks,
   relatedCollectionSlugs,
 } from './collections';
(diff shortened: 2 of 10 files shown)


strelov1:
Both findings looked at; one taken as proposed, one not.

**Footer links → `resolve('/collections/[slug]', …)` — taken.** Valid, and for a sharper reason than the base path: mine was the only internal link in that file not going through `resolve()`, and I had suppressed `svelte/no-navigation-without-resolve` to make it work — silencing the rule that exists to catch exactly this. `popularCollectionLinks` now returns slug + title and the component builds the URL, matching the six other dynamic-route call sites. The suppression is gone.

`paths.base` is empty today so there was no live bug, but `paths.relative` **is** on, which makes this worth verifying rather than assuming. Resolved links per page depth:

| Page | href emitted | resolves to |
|---|---|---|
| `/` | `./collections/aws` | `/collections/aws` |
| `/companies` | `./collections/aws` | `/collections/aws` |
| `/companies/stripe` | `../collections/aws` | `/collections/aws` |
| `/docs/api/jobs/search-get` | `../../../collections/aws` | `/collections/aws` |
| `/collections/python` | `../collections/aws` | `/collections/aws` |

Targets answer 200.

**Spy on `toLocaleString` to assert `'en-US'` — not taken.** That asserts the call rather than the metadata, and no other helper in this repo pins a locale that way, so it would make this one inconsistent for no behavioural gain.

The observation underneath it was right, though, and I fixed the real problem instead: `it('groups digits the same way regardless of locale')` asserted `toContain('1,234')`, which is not locale-independent — on a de-DE runner it fails just as the exact-output assertion above it does — and covered nothing that assertion did not. Dropped as redundant and misnamed. The remaining test now states what it does and does not catch: it fails the moment the pin is dropped on a runner whose locale groups differently, and would pass on an en-US one.

PR #2046, opened 2026-08-17
Rescue geography from eligibility statements beyond the US

strelov1:
## What

The description-based geography rescue in `jobderive` only knew US citizenship and
US security clearances. A posting whose location string parsed to nothing but
"Remote", and whose description said "you must have the right to work in the UK",
stayed in the `global` bucket and was served to everyone — the exact mislabelling
the rescue exists to prevent, just under a different flag.

This generalizes the single US predicate into a rule table: each rule pairs anchored
eligibility phrases with the geography they pin. Matching rules are **unioned**
rather than resolved by priority, because a posting open to two areas is genuinely
open to both. The gate is unchanged in spirit — it still only fires where the
location dictionary pinned nothing.

## Phrases came from the live catalogue, not from guesswork

A full-text sweep over production, verifying each hit actually contained the phrase
verbatim and counting how many carried no resolved geography:

| Area | Catalogue hits | Verbatim | Geo-unpinned |
|---|---|---|---|
| `authorized to work in the United States` | ~950 | 53/60 | **15** |
| `australian citizen` | 98 | 44/50 | **16** |
| `right to work in the UK` | 435 | 35/60 | 3 |
| `eligible to work in Canada` | 79 | 38/50 | 1 |
| `right to work in the EU` | 45 | **1**/45 | 0 |

Two findings shaped the result:

- **`authorized to work in the United States` was absent from the US list entirely**,
  despite being the single highest-volume eligibility phrase in the catalogue.
- **The EU list is deliberately thin.** EU-restricted postings overwhelmingly name a
  member state, which the location dictionary already pins.

Brazil, Germany, Singapore, Ireland, India, Mexico, NZ and Japan were measured and
deliberately left out: they have verbatim matches but **zero** geo-unpinned ones, so
a rule for them would only ever fire where the geography is already correct.

## A real bug fixed on the way

`without sponsorship` was suppressing its own signal. `without` is a negation word,
so `"authorized to work in the United States without the need for current or future
sponsorship"` — a phrasing quoted verbatim from production — read as a denial. That
tail *strengthens* an eligibility requirement, so it is now stripped before the
negation scan.

## Verification

- `go test ./...` — pass
- `go vet -tags=integration ./...` — pass
- `gofmt`, `go vet`, `govulncheck`, `golangci-lint` — clean
- Replayed against 437 unpinned production descriptions: 3 rescued (2x AU, 1x US)

## Deploy note

Dictionary changes do not reach stored rows on their own. This needs
`cmd/backfill-derive` followed by `cmd/reindex` after merge. The backfill runs
~15h on production — do not start it on a Friday.

## Summary by CodeRabbit

* **New Features**
  * Expanded work-authorization detection to recognize eligibility requirements for the US, EU, UK, Australia, and Canada.
  * Improved handling of combined locations and region-specific eligibility.
* **Bug Fixes**
  * Correctly interprets sponsorship wording and avoids treating valid eligibility statements as negations.
  * Improved detection for unpinned locations without applying an automatic US or North America designation.

### DIFF
--- a/internal/jobderive/jobderive.go
+++ b/internal/jobderive/jobderive.go
@@ -106,16 +106,19 @@ type Derived struct {
 // and skills (structured ∪ dictionary).
 func Derive(in Input) Derived {
 	geo := location.Parse(in.Location)
-	// Geography precedence: the location dictionary → a US-only description signal.
-	// When the location left the geography unpinned (no country, and either no region
-	// or only the bare-"Remote" global bucket) but the description states a hard
-	// US-only eligibility requirement, the job is US-restricted, not open-anywhere —
-	// pin it to the US so it leaves Global/Worldwide. Like the work-mode fallback
-	// below, prose is the lowest-priority source and only fills what the location
-	// dictionary left blank; a resolved place is never overridden.
+	// Geography precedence: the location dictionary → an eligibility signal in the
+	// description. When the location left the geography unpinned (no country, and
+	// either no region or only the bare-"Remote" global bucket) but the description
+	// states a hard eligibility requirement — US citizenship, EU work rights, a UK
+	// right to work — the job is restricted to that place, not open-anywhere, so pin
+	// it there and let it leave Global/Worldwide. Like the work-mode fallback below,
+	// prose is the lowest-priority source and only fills what the location dictionary
+	// left blank; a resolved place is never overridden.
 	countries, regions := geo.Countries, geo.Regions
-	if usOnly(countries, regions, in.Description) {
-		countries, regions = []string{"us"}, []string{"north_america"}
+	if geoUnpinned(countries, regions) {
+		if c, r := location.EligibilityFromDescription(in.Description); len(r) > 0 {
+			countries, regions = c, r
+		}
 	}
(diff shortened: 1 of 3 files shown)


strelov1:
Verified and fixed in 58ee73a0.

Confirmed the exact case: `EligibilityFromDescription("You must have the right to work in the Ukraine.")` returned `gb`/`uk` before the fix.

Matching is now whole-word, with one deliberate exception — a single trailing `s`, because the plural is how these statements are actually written ("United States citizens") while the lists carry the singular. Nothing longer counts, so `us citizen` no longer reaches into `US citizenship`; that is a separate entry matching itself exactly.

That strictness cost the accidental coverage a shorter entry used to give a longer one, so `australian citizenship`, `british citizenship` and `canadian citizenship` are now listed explicitly alongside their `citizen` forms — the "explicit variants" route you suggested.

Regression cases added for `Ukraine` (both that it no longer matches, and that the genuine UK sentence still does), plus the plural and the citizenship forms. Replayed against the same 437 unpinned production descriptions: coverage unchanged, same 3 postings rescued.

PR #2051, opened 2026-08-17
Let an unambiguous city state its own country

strelov1:
## The gap

A third of the catalogue carries no country and no region. Sampling 437 such
postings showed why — and it is not that they are worldwide-remote, which only 2%
of them are:

| Why a posting is unpinned | Share |
|---|---|
| Names a real place the parser discarded | **82%** |
| No location string at all | 14% |
| Genuinely says remote/anywhere | 2% |

"Colorado Springs", "Eden Prairie", "Benidorm", "Palma de Mallorca". The city
reached the cities facet; the geography facets stayed empty; the posting was
invisible to every country and region filter.

## Why it was discarded

The dictionary knew these places *and their country*. A previous review
deliberately refused to use it, because "most populous wins" would file a
Birmingham (UK) role under the US. That objection is correct — but it was
answered with a blanket ban, which buys silence on the ambiguous "Valencia" at
the price of silence on the unambiguous "Eden Prairie" too.

Across the dataset's aliases, **98% are claimed by exactly one country**.

## What this does

The ban narrows to what it was protecting. An alias claimed by more than one
country is marked `Contested` at load time and still states nothing; the rest
state their country. Two guards, both derived from measurement rather than
reasoning:

**1. The city is a last-resort signal.** It is applied after the whole line is
read, and only when nothing else stated a country. `Anna, Illinois, United
States` names a town in Russia *and* the state that actually places the job;
without this guard the city added Russia beside it. This alone took the change
from 9 wrong countries per 2000 postings to 3.

**2. `clinch`'s slug splitter preferred the shortest resolving window**, which
silently relied on junk fragments not resolving at all. "San" is a real town in
Mali, so "San Francisco" split in half. A window that is *itself* a city name now
wins outright, longest first.

## The counter-intuitive half

The second commit *widens* the dataset from pop>=15,000 to pop>=1,000 — and that
makes the parser **more** conservative, because coverage and caution move
together here. At 15,000 the only Taft in the file is Iranian, so the name looks
unambiguous and confidently mislabels a US posting. At 1,000 the Californian and
Filipino Tafts appear too, the name becomes contested, and it correctly says
nothing. Same for "Somerset", "San", "Young".

Taken naively that also breaks things, and it did: GeoNames at pop>=1,000
contains villages called "Engineer", "California" and "Attica". So the two uses
of the dataset are separated — a place below 15,000 is **evidence only**: it can
mark an alias contested, but never registers one. What the parser asserts is
unchanged; only its scepticism grew.

## Measured, not argued

Against the unmodified parser, on 2000 production postings:

| | commit 1 | commit 2 (final) |
|---|---|---|
| countries added where there were none | 53 | **39** |
| countries contradicted | 3 | **1** |

Fourteen fewer rescues bought back two of three wrong answers. For a dictionary
whose contract is "never guess", that is the right side of the trade.

Spot checks: `Ilminster, Somerset` (was US, now correctly silent), `Taft` (was
Iran, now silent), `ITA - Lazio - Roma` (was Paraguay, now silent),
`Colorado Springs`/`Benidorm`/`Eden Prairie` all correct.

`ParseResidence` benefits identically — a CV saying "Colorado Springs" now places
the candidate.

## Known residual

`Thomas Memorial Hospital (TMH)` resolves `gh` (Tema, Ghana) and `best` resolves
`nl` (Best, Netherlands). Both are single-claimant real places whose names double
as ordinary English words, so contest detection cannot reach them. A stop-list is
the right tool; neither appeared often enough in the sample to earn one yet.

## Cost

Almost none, after the last commit. Widening the dump was right for accuracy and
wrong for size — the extra 136k rows existed only to recover one bit per alias,
and that bit is known at generation time. `cmd/gen-cities` now counts claims
across the whole dump, marks contested aliases, and writes out only the places
large enough to state anything:

| | before this branch | mid-branch | final |
|---|---|---|---|
| embedded file | 3.3MB | 12MB | **3.3MB** |
| process-start parse | 44ms | 205ms | **49ms** |

That 205ms was paid by every process loading the package, `cmd/queue-metrics`
included, which runs once a minute.

Still needs `cmd/backfill-derive` then `cmd/reindex` after merge (~15h backfill).

## Verification

`go test ./...`, `go vet -tags=integration ./...`, `gofmt`, `govulncheck`,
`golangci-lint` — all clean.


strelov1:
Verified and fixed in ec6268b9.

Confirmed the order-dependence directly: `Parse("Recife, Benidorm")` returned `br` while `Parse("Benidorm, Recife")` returned `es` — same line, different answer depending on word order.

Took your approach: the candidate countries are collected and the fallback applies only when exactly one remains. Two cities in the *same* country still speak, since agreement is about the country rather than the name.

Regression tests assert both token orders explicitly (order-dependence being the actual bug), plus the agreeing case `Colorado Springs, Eden Prairie`.

Cost nothing measurable — re-run against the same 2000 production postings, the numbers are unchanged: 39 countries added, 1 contradicted. The case simply did not occur in the sample, which is exactly why it needed a test rather than a measurement.

PR #2072, opened 2026-08-17
Stop double-crawling iCIMS, recognise two more ATS, and add 475 harvested boards

strelov1:
Rebased onto #2073, which landed 2105 boards from the same aggregator while this was open. The two harvests sampled different slices — only 47 of these boards were already there — so the union stands at **475 new boards** here.

## 1. 37 iCIMS boards were crawled twice

The icims adapter reaches a board either way: a bare slug becomes the host `careers-<slug>.icims.com`, and a dotted board IS that host. `sources/icims.yml` carried 37 boards under both spellings, so every run crawled those sites twice and stored one row-set per spelling.

`boardDedupeKey` now folds the two forms together through a per-provider board-identity hook, so the loader collapses them even if a future harvest reintroduces one. Case folding alone could not see this.

Worth noting on the data merge: the bare rows carried the slug in their `company` field and the dotted rows the real display name — `vet` vs `Vet Jobs`, `rochebros` vs `Roche Bros`. Deleting the dotted rows naively would have discarded 33 real company names, so the surviving row takes the name.

Board ids are also trimmed at parse now. Whitespace makes a board that 404s — adapters paste the string into a URL — and it hid a duplicate: a harvested UKG board arrived with a trailing space and so did not collide with the same board already in the file.

## 2. atsboard recognises Avature and softgarden's regional career host

- **Avature** — the board is the career-site host, the shape `avature.yml` already stores (`deloittecm.avature.net`). Vanity hosts (`jobs.ea.com`) stay out, like every other custom-domain ATS here.
- **softgarden** — tenants also appear under `<tenant>.career.softgarden.de`. The label is the same board the adapter fetches at `<board>.softgarden.io` (verified live: `agilitaschweiz` answers on both), so only the apex differs.

Both widen the accept-set `internal/contribution` pays out on. Deliberate and narrow: both platforms are already crawled, both board ids are exactly what their board file stores, and neither is one of the shapes `atsdetect.FromURL` documents as excluded on purpose.

## 3. A Manatal prober, because the seed's company name is not the board's owner

`chooseCompany` falls back to the seed's company when the platform publishes none, and the seed's company is the posting's *hiring organisation*. Manatal is heavily used by recruitment agencies, so that is the agency's client: one client name got stamped on three different agency boards and four more read `confidential`. Both manatal and pageup label every posting with the entry's company, so a wrong label poisons the whole board.

`manatalProber` reads the tenant off the hosted career page's title — the board's owner, not whoever a posting is for. It also replaces the `adapterProber` fallback, so a candidate costs one request instead of a whole crawl: the 157-board run went from 3m22s to 40s.

PageUp deliberately gets no such prober — its listing titles are `Job Search`, `Current Opportunities`, `CSU Careers | CSU`, and guessing a name out of those is what the dict-only rule forbids. Its 19 boards were checked against their own listing pages one by one instead; five disagreed with the seed and take the platform's name. Two boards labelled ANSTO are in fact the Tasmanian Department of Education and the Tasmanian whole-of-government board.

## 4. 475 boards

Same play as #2070/#2073, using the existing `cmd/harvest-role` and `cmd/harvest-boards`: 20000 job pages → 3867 distinct boards → every new candidate probed against its own platform API → only those that answered with jobs.

The yield is lopsided, and usefully so: manatal, ukg, rippling, careerplug and oracle carried it, while Workday contributed 8 out of 1007 boards seen and Greenhouse 1 out of 212. That is what saturation looks like — the boards were at the thin providers. Avature and softgarden appear only because of change 2.

Also dropped: `taleo/AAR aarcorp.taleo.net/aar_mobile`, the mobile mirror of its `/2` careersection — a second careersection is a second `external_id` namespace over the same requisitions. Checked and kept: the UKG tenants holding several boards (MediLodge, Arbors of Ohio, Pharmascience) return disjoint opportunity sets, so those are per-facility boards.

## Checks

`gofmt -l`, `go build ./...`, `go vet ./...`, `go vet -tags=integration ./...`, `go test ./...` — all clean. No harvest run was rate-limited or aborted, and no board file carries a duplicate board under the folded identity.

### DIFF
--- a/cmd/harvest-boards/adapter_prober_test.go
+++ b/cmd/harvest-boards/adapter_prober_test.go
@@ -45,7 +45,7 @@ func TestProberForPrefersTheBespokeProber(t *testing.T) {
 func TestProberForFallsBackToTheProvidersAdapter(t *testing.T) {
 	// Platforms with a board-keyed adapter and no bespoke prober. Before the fallback these
 	// were unharvestable: harvest-boards refused the provider outright.
-	for _, provider := range []string{"rippling", "zohorecruit", "ukg", "manatal", "phenom", "hibob"} {
+	for _, provider := range []string{"rippling", "zohorecruit", "ukg", "phenom", "hibob"} {
 		p, ok := proberFor(provider)
 		if !ok {
 			t.Errorf("proberFor(%q) not found; the provider has an adapter and is board-keyed", provider)
--- a/cmd/harvest-boards/manatal_prober.go
+++ b/cmd/harvest-boards/manatal_prober.go
@@ -1,0 +1,80 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"strings"
+
+	"golang.org/x/net/html"
+)
+
+// manatalProber validates a Manatal board "<tenant>" by counting the postings linked from its
+// hosted career page ("careers-page.com/<tenant>") and reading the tenant's own name off that
+// page's title.
+//
+// The name matters more here than on most platforms. Manatal is heavily used by recruitment
+// AGENCIES, so a posting's hiring company is routinely NOT the board's owner — a seed built
+// from an aggregator's hiringOrganization labelled three different agency boards with the same
+// client, and the adapter stamps every posting with the entry's company. The career page names
+// the tenant that owns the board, which is the only correct label.
+//
+// Without this the provider fell through to adapterProber, which runs a whole crawl per
+// candidate and publishes no name at all.
+type manatalProber struct{}
+
+// manatalJobLink captures a posting's id from a Manatal career-page job link
+// (/<tenant>/job/<id>), so the title and apply links to one posting count once.
+var manatalJobLink = regexp.MustCompile(`/job/([A-Za-z0-9]+)`)
+
+// manatalTitleSuffix is what Manatal appends to every hosted career page's title. The tenant
+// name is whatever precedes it.
+const manatalTitleSuffix = "| Career Page"
+
+func (manatalProber) probe(ctx context.Context, c httpClient, board string) (string, int, error) {
+	root, err := c.GetHTML(ctx, fmt.Sprintf("https://careers-page.com/%s", board))
+	if err != nil {
+		return "", 0, nil
+	}
+	ids := map[string]bool{}
+	var title string
+	var walk func(*html.Node)
+	walk = func(n *html.Node) {
+		if n.Type == html.ElementNode {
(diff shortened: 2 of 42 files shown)


strelov1:
Worked through the review. Two findings held, one did not, and one pointed at a bigger problem than it described.

**UKG duplicates — both real, one worse than reported.**

- `Atlantic Union Bank` did arrive with a trailing space and so slipped past `boardDedupeKey`. Fixed at the cause: board ids are now trimmed in `ParseRawEntries`, because whitespace makes a board that 404s — the adapters paste the string straight into a URL.
- `Mahaska Health` is not two live boards. The pre-existing spelling is `.../MAH1000MAHC/JobBoard/<guid>`, and `splitUKGBoard` requires exactly `host/tenant/guid`, so that entry has never crawled. The harvested three-part spelling replaces it.

**PageUp ANSTO 1035/759 — not a duplicate.** I fetched both listings: 15 job ids against 25, zero overlap.

But the comment landed on something real. Neither board is ANSTO: `1035` is `Current Opportunities - Department of Education Tasmania` and `759` is `Tasmanian Government Jobs`. So I checked all 19 new PageUp boards against their own listing pages — five disagreed with the seed and now carry the platform's name.

**Manatal names — right, and structural.** `chooseCompany` falls back to the seed's company when the platform publishes none, and the seed's company is the posting's *hiring organisation*. Manatal is heavily used by recruitment agencies, so that is the agency's client: one client name was stamped on three different agency boards, and four more read `confidential`. Since the adapter labels every posting with the entry's company, a wrong label poisons the whole board.

Rather than patch the names, Manatal now has a prober of its own that reads the tenant off the hosted career page title — the board's owner rather than whoever a posting is for. It also replaces the `adapterProber` fallback, so a candidate costs one request instead of a whole crawl: the 157-board run went from 3m22s to 40s.

PageUp deliberately gets no such prober. Its listing titles are `Job Search`, `Current Opportunities`, `CSU Careers | CSU` — extracting a name from those is guessing, which the dict-only rule forbids. Hence the one-by-one check above.

**Also dropped:** `taleo/AAR aarcorp.taleo.net/aar_mobile`, the mobile mirror of its `/2` careersection — a second careersection is a second `external_id` namespace over the same requisitions.

**Checked and kept:** the UKG tenants holding several boards (MediLodge, Arbors of Ohio, Pharmascience). I queried three MediLodge boards on the shared tenant and they return disjoint opportunity sets — per-facility boards, not duplicates.

PR #2075, opened 2026-08-17
Say which query params the search ignored

strelov1:
## Why

A filter param the API does not read is dropped in silence. That is what made `country=it` — the facet is `countries` — return the unfiltered catalogue and read as a real answer about Italy. Measured on prod while diagnosing a user report:

| query | total |
|---|---|
| `countries=it&employment_type=contract` | 1 119 |
| `country=it&employment_type=contract` | 165 645 |

The second is the whole contract catalogue. Nothing in the response said so.

## What

`meta.ignored_params` on `/jobs/search` and `/agent/jobs/search`, with `did_you_mean` when the caller only got the grammatical number wrong (most facets are plural). Omitted entirely when there is nothing to report, so an ordinary response carries no field readers learn to skip.

Ignored rather than refused: saved searches and shared links still carry retired params, and a 400 would break every one of them.

The suggestion is number-only, deliberately — no edit distance. `country` → `countries` needs a threshold loose enough to also invent suggestions for params that were never a typo.

## Docs

The public filter docs claimed every facet ANDs. Geography is the exception — `regions`, `countries` and `cities` OR into one group, so adding `countries=IT` to `regions=eu` widens nothing. That is the trap behind the original report. Stated, with the one-country recipe.

## Tests

`go vet`, `go vet -tags=integration`, full unit suite, golangci-lint clean. `TestScalarFilters_EachOneStillNarrowsAQuery` guards the hand-kept filter list against rotting into a lie — verified it fails when a decoy entry is added.

## Summary by CodeRabbit

* **New Features**
  * Search, company, facet, and market coverage responses now identify unsupported query parameters in `meta.ignored_params`.
  * Suggestions may be provided for likely parameter-name corrections.
  * Clean responses retain their existing format without unnecessary metadata.
  * Geography filters are documented as shared OR groups, while other facets remain combined with AND logic.

* **Documentation**
  * Added guidance to check ignored parameters, which may otherwise broaden search results.
  * Added an example for filtering by country without specifying a region.


strelov1:
@coderabbitai review

PR #2116, opened 2026-08-18
Choose a spelling by how the employer writes its name, and stop backfill-derive undoing the merges

strelov1:
Fourth and last of the election fixes the prod dry runs have surfaced.

## Why

In the `--min-jobs 100` wave, job count elects a **squashed** canonical over a hyphenated one **73 times out of 162** spelling merges.

Some are right — AT&T really is one word, so `att` beats `at-t`; so do `autonation`, `aspentech`, `agileengine`. Others are plainly wrong: `accenturefederalservices`, `acehardware`, `westerndigital`, `allegisglobalsolutions`.

## The signal is the name, not the slug

An employer that writes its name in several words should key as several words.

This is also why the obvious version of the rule failed before. "Prefer the more hyphenated slug" reads the slug — and in a slug an apostrophe is indistinguishable from a space, so it elects `domino-s`(1 job) over `dominos`(14,396) and `al-fa-bank`(20) over `alfa-bank`(1,617).

Reading the name fixes both, because the preference **only speaks when it discriminates**:

| Group | Names | Multi-word? | Outcome |
|---|---|---|---|
| Domino's | "Dominos", "Domino's" | neither | count → `dominos` ✓ |
| Alfa Bank | "Alfa Bank", "Al Fa Bank" | both | count → `alfa-bank` ✓ |
| Western Digital | "WesternDigital", "Western Digital" | one | → `western-digital` ✓ |
| AT&T | "AT&T" | neither | count → `att` ✓ |

Trailing legal forms do not count as words: "Ace Hardware Corporation" is a two-word employer.

## Note

Wave 1 is already applied and its canons are **frozen**, so `turnertownsend` and `nexthire` keep theirs — freezing beats tidiness once a url has been redirecting. Both were minted about an hour ago and could be reversed cheaply if you'd rather; say so and I will.

`go test ./...` 165 ok · `go vet -tags=integration` clean.

## Summary by CodeRabbit

- **Improvements**
  - Improved company merge planning to prefer candidate names that preserve distinct multi-word naming.
  - Maintained job-count ordering when names provide no clear distinction.
  - Improved fallback handling when a reproducible company identifier is unavailable.
- **Documentation**
  - Updated company identity guidance to clarify name-based selection and tie-breaking behavior.

### DIFF
--- a/cmd/backfill-derive/derive_alias_test.go
+++ b/cmd/backfill-derive/derive_alias_test.go
@@ -1,0 +1,67 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/strelov1/freehire/internal/db"
+	"github.com/strelov1/freehire/internal/jobhash"
+)
+
+// TestDeriveRow_ResolvesThroughTheAliasRegistry guards the merges against this worker.
+//
+// deriveRow re-derives every column from jobderive, which is a PURE function with no
+// knowledge of company_slug_aliases. Left alone it would rewrite every posting a merge moved
+// back to the spelling its source happened to use — silently undoing the whole spelling class,
+// and churning role_fingerprint with it, since the fingerprint is computed from the company
+// slug. A backfill is a routine ~15h run, so this is not a hypothetical.
+func TestDeriveRow_ResolvesThroughTheAliasRegistry(t *testing.T) {
+	canon := map[string]string{"dollartree": "dollar-tree"}
+	job := db.Job{
+		ID: 1, Title: "Backend Engineer", Company: "DollarTree",
+		Source: "adzuna", ExternalID: "1", CompanySlug: "dollar-tree",
+	}
+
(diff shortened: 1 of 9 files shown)


strelov1:
## Added: backfill-derive would have undone every spelling merge

Found while checking what else re-derives `company_slug`.

`deriveRow` computes it from `jobderive`, which is **pure** and knows nothing of `company_slug_aliases` — so for a merged posting it yields the spelling the *source* used, i.e. the retired one, and the worker writes it. `role_fingerprint` is computed from that slug, so every merged posting's repost identity would churn on the way out too.

Not hypothetical: `backfill-derive` is a real ~15h run, and nothing about it would have looked wrong.

The registry is now loaded **once per run, before any worker starts**, so every row resolves against the same view. A run that cannot read it **fails** rather than proceeding — an empty registry is indistinguishable from a catalogue with no merges, which is exactly the state that would make the damage invisible.

PR #2170, opened 2026-08-21
Say the header scope in glyphs, not in words it already drew

strelov1:
## What and why

The header's scope trigger read `🏠 Remote · Worldwide +3`. The house already says
"remote", and "Worldwide" is a long word for a fact a globe states in one glyph. Both
collapse into an icon cluster — work format, then a globe when the worldwide region
leads the geography — lapped over each other the way `CountryFlagStack` laps flags:
negative margin, a ring in the box's own background, earlier icon on top. What stays
in words is the part that needs them: the head country/region and its `+N`.

`summarizeScope` now returns what to draw (`icons`, `text`, `extra`) instead of a
pre-joined string. The worded summary survives as `label`, which the button wears as
its accessible name **and** its `title` — a glyph reads to no screen reader, and the
tooltip is now the only place the full scope is spelled out.

Two judgement calls worth a look:

- **`Location` stays** as the text whenever no geography is picked, including when a
  work format is (`🏠 Location`). A lone glyph with a chevron does not read as
  something you can open.
- **The globe stands in only for a region-shaped param** carrying `global`. A city
  literally named "global" is a place, not everywhere, and keeps its name. Covered by
  a test.

| Selected | Before | After |
|---|---|---|
| nothing | 🌐 Location | 🌐 Location |
| Remote | 🏠 Remote | 🏠 Location |
| Remote + Worldwide + 3 more | 🏠 Remote · Worldwide +3 | 🏠🌐 +3 (lapped) |
| Remote + Europe + 1 more | 🏠 Remote · Europe +1 | 🏠 Europe +1 |

Verified in a browser at all four states; the first lap value (`0.35em`) ate the
globe's left edge, `0.2em` keeps both glyphs legible.

## Checklist

- [ ] `go build ./...`, `go vet ./...`, and `gofmt -l .` (prints nothing) pass. — n/a, no Go changed.
- [ ] `go test ./...` passes (and `go test -tags=integration ./...` if I touched DB/handlers). — n/a.
- [x] I regenerated committed artifacts when their source changed — n/a, no SQL or contract source changed.
- [ ] For `design-system/` changes: `pnpm run check` and `pnpm run build` pass. — n/a, none.
- [x] For `web/` changes: `pnpm run check` (0 errors) and `pnpm run build` pass; `pnpm vitest run` — 1144 tests green.
- [x] This stays within freehire's core/extension boundary — one component and its pure summarizer.

## Summary by CodeRabbit

- **New Features**
  - Location filters now display multiple selected location icons in an overlapping stack.
  - Additional selected locations are summarized with a count.
  - Filter labels adapt responsively and remain accessible with descriptive text.
  - Worldwide selections are represented visually without cluttering the visible summary.

- **Bug Fixes**
  - Improved location summary handling for combinations of worldwide, country, and city selections.

### DIFF
--- a/web/src/lib/components/HeaderLocationFilter.svelte
+++ b/web/src/lib/components/HeaderLocationFilter.svelte
@@ -33,10 +33,14 @@
   const summary = $derived(
     store
       ? summarizeScope(store, variant === 'companies' ? COMPANIES_SCOPE : JOBS_SCOPE)
-      : { icon: 'globe' as const, label: 'Location' },
+      : { icons: ['globe' as const], text: 'Location', extra: 0, label: 'Location' },
   );
   const ICONS = { globe: Globe, remote: House, hybrid: Blend, onsite: Building } as const;
-  const TriggerIcon = $derived(ICONS[summary.icon]);
+  // The head geography and its roll-up as one string; either half can be empty (a
+  // format with no geography draws its icon alone, worldwide draws the globe alone).
+  const trailing = $derived(
+    [summary.text, summary.extra > 0 ? `+${summary.extra}` : ''].filter(Boolean).join(' '),
+  );
 
   // Facets "Clear all" resets, per mode.
   const CLEAR_PARAMS = {
@@ -105,6 +109,7 @@
     aria-haspopup="menu"
     aria-expanded={open}
     aria-label={summary.label}
+    title={summary.label}
     onclick={(e) => {
       // Stop the toggle's own click from reaching the window outside-handler, which
       // would otherwise immediately re-close the just-opened popover.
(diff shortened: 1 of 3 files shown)


strelov1:
## CodeRabbit autofix review

Reviewed one CodeRabbit thread and applied no code change from it. Reason below.

**"Detect worldwide across all selected geography values"** — declined. The trigger has always
named the *head* geography and counted the rest: `regions=eu,uk` reads "Europe +1" and never
names UK. Singling out `global` for a second mention would make `regions=eu,global` read
"Europe +1 · Worldwide" — counted in the `+1` and spelled out beside it. Whether the globe
should override which geography leads is a design question about the roll-up, not a defect in
this change, and it is not one this PR opens.

Separately, CI's design-system token ratchet caught a real regression here: `-ml-[var(--lap)]`
pushed the file's arbitrary-Tailwind-value count from 2 to 3. Fixed in 99be46d0 with a plain
spacing utility, no baseline edit. The lap dropped to 2px on the way — 4px covered the globe's
left third in the browser.

PR #2172, opened 2026-08-21
Open a first-time visitor on the jobs near them

strelov1:
## What and why

The front page is every job on earth, ordered by freshness, most of them somewhere the
reader cannot work. The one fact the site can infer without asking is where they are —
Cloudflare names the country on every request for free. That becomes the opening scope:
their macro-region **plus worldwide**, so a remote-everywhere posting is never hidden by
a regional default.

Applies to the standalone jobs feed only, once per browser, and only when nothing else
said what geography to show.

### The cache constraint drove the shape

The region is served from **its own endpoint marked `private, no-store`**, never from
page data. My first design put it on `event.locals` and returned it from the root
layout, the way `locale` is handled — that was wrong: server-load data is serialized
into the document, and that document is edge-cached by URL alone, so the HTML would
have differed by country. `locale` gets away with it because it comes from a cookie the
visitor's own request carries.

Verified by diffing whole documents for BR and US, two fetches each. They differ in
exactly two places — the per-request CSP nonce and a ticking "N seconds ago" — and both
also differ between two fetches from the *same* country. Normalize those and all four
hash identically.

### Precedence, and the clear that must stick

1. **Any param in the URL** suppresses the guess — not only a geographic one. A link
   built to mean "senior jobs" must show its recipient what its sender saw.
2. **A stored filter set** wins, on the strength of being theirs rather than on what it
   contains.
3. **The marker** means this browser was already asked.

The marker is its own storage key, and that is the point: clearing the filters removes
`hire.jobFilters` outright, so a guess keyed on "storage is empty" would undo the clear
on every later visit. Only wiping browser storage re-arms it. Unreachable storage reads
as "already offered" — a guess that cannot be recorded is one that cannot be dismissed.

The scope is written by shallow `replaceState` and re-read rather than through
`filters.apply()`: on the standalone list an explicit write mirrors itself into storage,
and storage records what the visitor *chose*.

### Crawlers

Answered with no region. Most traffic here is automated and a rendering crawler would
otherwise index a feed scoped to its exit address. The **Lighthouse watchdog is
deliberately not treated as a crawler** — it has to walk the path visitors walk, or it
measures a page nobody is served.

### The measurement, including the two wrong answers

Replacing the list after hydration is measurable, so it was measured against real prod
data with the country pinned by header — and **the cost is a range, not a number**. The
same build read CLS **0.026** in one session and **0.246** an hour later, each stable to
four decimals across runs, against a 0.0014 baseline both times. LCP was unmoved in both.

Stashing every edit made in between and re-measuring reproduced 0.246 exactly, which
rules out the code: holding the outgoing rows removes the *collapse*, but not the
difference in height between twenty unscoped rows and twenty scoped ones — and that
difference is whatever the catalogue happens to be serving that hour.

Shipping automatic application is a deliberate call taken with that range in hand. The
checks that can overturn it are the field ones — the scheduled Lighthouse watchdog and
CrUX — with the suggestion form (no swap, no shift) as the recorded fallback.

Two earlier measurements were wrong outright and both are worth recording:

- **0.50** — measured against a built `adapter-node` server, which does **not** proxy
  `/api` (only the Vite dev server does). `API_INTERNAL_URL` fixes the server render and
  nothing else, so the SSR list arrived and the client's reload 404'd. It was measuring
  an error message replacing a list.
- **0.87** — working harness, but the hold-over was broken twice: its release effect
  tracked its own write and cleared the rows in the same tick that captured them, and
  the empty-state branch read `jobs.items` directly, drawing "No matching jobs" over a
  list that was merely reloading. Every row read now goes through one `displayItems`.

The tell both times was the shift attribution: a `footer` going from height 0 to 386 is
a page collapsing, which no ordinary list swap does.

### Review findings fixed

- **The offer could land on a page it was never made for.** `fetchRegion()` awaits, and
  the visitor can navigate away in that time onto a route that also has an empty query
  string. The pathname is captured before the request and every guard re-read after it.
- **A failed marker write re-armed the guess forever.** A quota-exhausted store answers
  `getItem` and throws on `setItem`; the marker never persisted and the guess returned
  on every visit, including to someone who had just cleared it. `markGeoScopeOffered`
  now reports whether the write stuck, and an unrecorded offer is not made.
- Two self-contradictions in the documents: the proposal claimed "no new endpoint" while
  defining one, and the spec said "no geography params" where the code suppresses on any
  param at all — the stricter rule is the correct one.

### Not done — ops, in another repo

The zone's IP Geolocation toggle and the nginx `CF-IPCountry` forward live in
`freehire-ops`. Until both land no header arrives and **the feature is inert by design
rather than broken** — worth knowing before anyone debugs the client code. Read the zone
before applying: a ruleset `PUT` replaces rather than merges.

## Checklist

- [ ] `go build ./...`, `go vet ./...`, `gofmt -l .` — n/a, no Go changed.
- [ ] `go test ./...` — n/a.
- [x] I regenerated committed artifacts when their source changed — n/a, no SQL or contract source changed.
- [ ] For `design-system/` changes — n/a, none.
- [x] For `web/` changes: `pnpm run check` (0 errors), `pnpm run build`, `pnpm vitest run` (1169 passed), eslint, and the design-system token ratchet all pass.
- [x] This stays within freehire's core/extension boundary.

## Summary by CodeRabbit

* **New Features**
  * Standalone job searches can receive a one-time location scope based on the visitor’s region.
  * Inferred locations include the visitor’s region and worldwide listings without altering other job or company views.
  * Location controls indicate when a scope was inferred and provide accessible options to edit or clear it.
  * Explicit URL selections and saved filters take precedence over inferred locations.

* **Bug Fixes**
  * Preserved feed content and layout while an inferred location scope loads.
  * Prevented crawlers and unavailable location data from receiving inferred scopes.

* **Tests**
  * Added coverage for country mapping, crawler handling, filter precedence, persistence, and storage failures.


strelov1:
## Review findings

Six threads, four acted on.

**Fixed — the offer could land on a page it was never made for.** Right, and a real bug.
`fetchRegion()` awaits, the visitor can leave in that time, and every route they might
land on also has an empty query string, so the guards passed and the scope was written
onto whatever `location.pathname` had become. The pathname is now captured before the
request and every guard re-read after it.

**Fixed — a failed marker write re-armed the guess forever.** Also right, and worse than
it reads: reads and writes fail independently, so a quota-exhausted store answers
`getItem` and throws on `setItem`. The marker never persisted, the guess came back on
every visit, and it came back to exactly the person who had just cleared it — the one
failure the separate key exists to prevent. `markGeoScopeOffered` now reports whether
the write stuck, and an unrecorded offer is not made at all. Covered by a test with a
working reader and a throwing writer.

**Fixed — the proposal claimed "no new endpoint" while defining one.** Correct, my
document contradicted itself.

**Fixed — "no geography params" vs "no URL parameters".** Correct, and the stricter rule
is the right one: a link built to mean "senior jobs" must show its recipient what its
sender saw. The spec now says what the code does.

**On the CLS gate — you were right, and for a better reason than the one given.** The
finding was that a good absolute score does not satisfy a no-regression rule. What
re-measuring found is worse: **the figure is not reproducible.** The same build read
0.026 in one session and 0.246 an hour later, each stable to four decimals. Stashing
every edit made in between reproduced 0.246 exactly, which rules out the code — holding
the outgoing rows removes the collapse but not the height difference between twenty
unscoped rows and twenty scoped ones, and that difference is whatever the catalogue is
serving. The spec, the design notes, the task record and the PR body now state the range
instead of the flattering end of it. Shipping automatic application stands as a
deliberate decision taken with that range in hand; the lab threshold is replaced by the
checks that can actually overturn it (the Lighthouse watchdog, CrUX) with the suggestion
form as the recorded fallback.

**Not acted on — "complete the edge configuration before release".** Agreed and already
tracked as tasks 7.1-7.4; the config lives in a separate ops repo, so it cannot land in
this PR. Called out in the PR body because the failure mode is silence: with no header
the feature is inert by design and looks broken.

PR #2180, opened 2026-08-21
Hold a DSDS entity's props to the ones the component takes

strelov1:
`validate-docs` already compares two fields of a DSDS entity back against the package — the files it points at, and its token list, both ways. The prop list was the third copy and nothing reconciled it.

`Button` grew `target` and `rel` in #1920 and the entity did not, so the reverse-tabnabbing default the component fills in for a `_blank` target was written down only in a source comment. That prop list is the half of an entity an agent reads to decide what it may pass, so a prop missing from it is a capability nothing can discover.

**What's here**

- `propsFrom(source)` — reads the destructure `$props()` fills, naming each prop the way a call site writes it: `class`, not `className`, and `...rest` kept whole so the passthrough has somewhere to be described. Balanced-brace walk rather than a regex — the inline type annotation between the destructure and `$props()` is a second brace block whose members are types, and an arrow type inside it puts an `=` where scanning for one would stop too early.
- The comparison, both ways, next to the token one.
- 7 tests in `scripts/validate-docs.test.mjs`, matching the sibling scripts' export-a-pure-function shape. `validate-docs.mjs` grew the `import.meta.url` main guard the siblings have, so importing it no longer runs the validation.
- Button's entity documents `target`, `rel` and `...rest`.
- `docs/verification.md` records what the check holds and how prop names are spelled.

**Evidence**

The check found exactly one drifted entity out of 26 before the docs fix:

```
✗ components.json: entity "component.button" does not document prop(s) target, rel, ...rest
```

After: `pnpm validate:docs` clean, `pnpm test` 19 files / 138 tests green, `pnpm lint` adds no new warnings.

### DIFF
--- a/design-system/docs/dsds/components.json
+++ b/design-system/docs/dsds/components.json
@@ -41,6 +41,18 @@
           "description": "If set, renders an <a> instead of a <button>."
         },
         {
+          "name": "target",
+          "type": "string",
+          "default": "undefined",
+          "description": "Anchor target; only meaningful alongside href."
+        },
+        {
+          "name": "rel",
+          "type": "string",
+          "default": "undefined",
+          "description": "Left unset, target=\"_blank\" still gets rel=\"noopener noreferrer\" — the component fills the reverse-tabnabbing default in so no call site has to. Passing rel yourself replaces it, including that default."
+        },
+        {
           "name": "class",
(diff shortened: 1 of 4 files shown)


gearsandcode (approved):
LGTM

PR #2221, opened 2026-08-27
Stop deleting the skill embedder right after declaring it

strelov1:
Follow-up to #2220, found while verifying the deployment path rather than by a test.

## The bug

`EnsureIndex` and `Rebuild.Prepare` each applied the index settings and then called
`ResetEmbedders`. Now that the settings declare the skill embedder, that deleted it in
the same call.

A full rebuild — the only way the embedder reaches production — would have produced an
index with **no** embedder. Every `sort=match` query would then 400, which this
package's error mapping turns into a 500. Meanwhile `facetSettings()` still declared
the embedder and every unit test still passed. Nothing would have logged.

## The fix

The reset is still necessary. Meilisearch merges settings, and merges `embedders` **by
key**: an update naming one embedder leaves any other in place, so an index that a
prior version gave a model-backed embedder would keep it — and keep embedding every
document — through any update that does not mention it.

What was wrong is the order. The reset moves into `ensure()`, **before** the settings
update, so the live embedder set ends up as exactly what the settings declare: nothing
inherited, nothing dropped. Both call sites lose their trailing reset.

## Tests

The gap was structural: a unit test over `facetSettings()` asserts what we *ask for*,
and only a live engine shows what the index *ends up with*. That is precisely where the
two diverged, so the new coverage is integration-tagged:

- the embedder survives `EnsureIndex`, at the declared width
- the embedder survives `Rebuild.Prepare` — the path that was broken
- a vector search ranks end to end against a real engine
- a job that loses its skills has its stored vector cleared, and stays findable by text

## Verification

`gofmt` clean, `go build ./...`, `go vet ./...`, `go vet -tags=integration ./...`,
`go test ./...`, `golangci-lint` clean on changed files. Integration tests run in CI —
Docker would not start locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed skill-based search setup so the configured skill embedder remains available after index creation and rebuilds.
  * Improved vector search ranking for skill matches.
  * Ensured stored skill vectors are cleared when a job’s skills are removed or matching data is unavailable, while keeping jobs searchable by keyword.
  * Preserved indexable documents when skill-matching weights cannot be loaded.

* **Tests**
  * Added coverage for embedder availability, vector ranking, skill removal, and unavailable matching data.

### DIFF
--- a/cmd/reindex/skillvector_test.go
+++ b/cmd/reindex/skillvector_test.go
@@ -32,8 +32,10 @@ func TestSplitJobs_CarriesTheSkillVector(t *testing.T) {
 }
 
 // A rebuild started before the rarity rollup ever ran must still rebuild. Losing the
-// match sort for a cycle is far better than not rebuilding the index at all.
-func TestSplitJobs_WithoutWeightsStillBuildsDocuments(t *testing.T) {
+// match sort for a cycle is far better than not rebuilding the index at all — and the
+// documents must stay ACCEPTABLE to the engine, which rejects any that omits the
+// vector field once the embedder is declared.
+func TestSplitJobs_WithoutWeightsStillBuildsIndexableDocuments(t *testing.T) {
 	job := db.Job{
 		ID: 1, Title: "Go Engineer", PublicSlug: "go-x", Category: "backend",
 		Description: "<p>Build things.</p>", Skills: []string{"go"},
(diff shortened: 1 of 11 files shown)


strelov1:
All four taken, one of them sharpened.

**The embedder-less window** — right, and it is a correctness point rather than a
latency one: a `sort=match` query arriving between the reset and the settings gets a
400. Fixed as suggested. Meilisearch's one-serial-queue-per-index guarantee is what
makes it safe, and this repo already relies on it elsewhere (`internal/search/search/AGENTS.md`
opens with it), so the ordering holds without the await.

**The clearing test** — a fair hit. It asserted the job stayed findable by text but
never that the vector was actually gone, which is the entire claim. It now re-runs the
vector search and requires the job to be absent from it.

**The stale no-weights contract** — the sharpest of the four. The spec described the
abandoned three-state design in one requirement and the current one in another, and an
indexer following the older text would omit the field and have its documents rejected.
Rewritten to name the two layers explicitly: construction distinguishes *why* there is
no vector; serialisation cannot, because a declared embedder rejects an omission.

**The stale comment** — corrected, along with the precondition just above it, which
tested `len(emb) == 0` and would now pass for the wrong reason.

PR #2236, opened 2026-08-31
Scan for secrets, lint new migrations, catch sqlc drift — and fix three checks that could not fire

strelov1:
Four checks the repository did not have, plus two glob traps that had silently
disabled three it thought it did. Each commit is one tool and the fixes it turned up,
and each was verified by planting the defect it exists to catch — not by counting
findings, since "the scanner is blind" and "the tree is clean" both print zero.

## Secrets — `gitleaks`

Nothing here looked for a credential in a commit. Formatting, vetting, linting and
`govulncheck` all pass over a key pasted into a config or a fixture.

Two moments, two controls. The **hook** reads the staged index and is the last point at
which a leak has not happened yet; it fails hard on a missing binary, because a scanner
that quietly skips makes a commit look examined. The **workflow** scans the whole history,
cannot be bypassed, and covers whoever never ran `lefthook install` — but by the time it
speaks the credential is on the remote, in every clone and in the runner's cache, and the
answer is to revoke it.

Ten findings on adoption, none a live credential: eight test fixtures, the extension's
Chrome manifest key (an RSA *public* key — publishing it is the mechanism, since the
extension id is derived from it), and a `logo.dev` publishable browser token the code no
longer carries. Each is allowlisted in `.gitleaks.toml` by its **line shape, never its
path**: gitleaks ORs `paths` with `regexes` and prunes the file before reading it, so a
path entry does not narrow a rule — it switches the scanner off for that file. `_test.go`
both invites such an entry and is where a live token would sit as a "realistic" fixture.

The workflow **asserts the commit count**. Given a repository it cannot read, gitleaks
prints `0 commits scanned`, says `no leaks found` and exits 0.

## Migrations — `squawk`

"Never edit an applied migration" was prose in AGENTS.md read by nothing, and nothing
looked at what a new migration does to a live table. That is the expensive class of
mistake: a `UNIQUE` constraint holds `ACCESS EXCLUSIVE` for the whole index build, a
constraint without `NOT VALID` scans the table and blocks writes, and Postgres queues lock
requests — so an `ALTER` waiting on a lock sits ahead of every reader arriving after it.
That is how a bounded migration took out signed-in profile reads on 2026-07-30.

Two facts about this repository shape it, and both are why a bare `squawk migrations/*.sql`
would be the wrong tool.

**Transactionality is a property of the FILE.** The runner wraps each migration unless its
leading comment carries `migrate: no-transaction`, and 18 of 156 do. Squawk has one global
flag, so `scripts/check-migrations.mjs` splits the invocation and reads the marker out of
the runner rather than copying it. Setting the flag globally is a lie either way: `true`
hides `prefer-robust-stmts` on exactly the files that can half-apply; `false` demands
`CREATE INDEX CONCURRENTLY` on files where it cannot run at all.

**The history stays.** Rewriting an applied migration is a worse hazard than any finding in
one, so the check reads only what a change ADDS. Unusually clean as ratchets go: unlike a
line of Go, a migration file genuinely never changes after it lands — no baseline file to
drift, no blame-by-line.

Two rules excluded, each on a fact rather than on taste — `require-lock-timeout` because
the runner sets it once per connection, session-level, and a copy inside every migration is
a copy that can drift; `require-statement-timeout` because it would abort a long index
build partway, and an aborted `CONCURRENTLY` build leaves an INVALID index behind (0117 and
0118 are the repair pair for exactly that). Those two corrections take the raw **1322
findings to 267** — four fifths of the noise was the tool being told the wrong thing about
how migrations run here.

Suppress a rule with `-- squawk-ignore <rule-name>` beside the statement, where the argument
sits next to the thing it argues about.

## Generated code — sqlc drift

`internal/platform/db` is generated and committed, and nothing compared the two halves. A
query edited without `make sqlc` ships the previous Go and every check stays green, because
each of them reads the committed file rather than the query it came from. The design-system
`check:dist` step exists for this shape of gap on the CSS side.

The two halves ask different questions. CI asks "does it differ from HEAD", and a bare
`git status --porcelain` answers that correctly on a fresh clone. The hook has to ask "did
regenerating leave anything UNSTAGED" — porcelain reports two columns, so the bare form
would count the staged generated files being committed as drift and block the very commit
that fixes it. Hence `git diff` plus `git ls-files --others`, and the second half is not
redundant: a NEW generated file is untracked, not modified.

Both invoke `make sqlc`, so `SQLC_VERSION` stays the only pin. A second pin would be a
second answer, and the drift between them would look exactly like stale code.

## Two glob traps that had disabled three existing checks

Both read as `skip, no matching staged files` — which looks like nothing to check, not like
a check that cannot fire. Measured with lefthook 1.13.6.

**`root:` discards before it globs.** It does not merely set the working directory: lefthook
first drops every staged file outside it, then applies the glob. `design-system-tokens` and
`design-system-adoption` both declare a radius over `web/src` — the comment above them said
so — while carrying `root: design-system/`, which made that half unreachable. `--file
web/src/lib/logo.ts` skipped both. They now run via `pnpm --dir` from the repository root;
not `--filter`, since these are separate packages with their own lockfiles rather than a
pnpm workspace.

**`**/` requires a separator.** `design-system/src/**/*.svelte` matches
`src/email/email-preview.svelte` and **not** `src/alert.svelte` — and every primitive is a
direct child of `src/`, so that pattern covered 1 of 26. It is the pattern `tokens` and
`docs` were globbed on.

Net effect: the token-coverage check — which exists to catch a hex literal or an arbitrary
Tailwind value the theme cannot move and the `.dark` selector cannot override — was
reachable from neither of the two places such a literal gets written.

## Verification

Each check was proved by planting its defect, and one probe failed usefully enough to keep:
gitleaks silently ignores `AKIAIOSFODNN7EXAMPLE` for its stop word, and a bare AWS access
key id is not a secret to begin with, so the first attempt looked exactly like a working
scan of a clean tree.

- **gitleaks** — history clean through the config; a planted secret reported in both `dir`
  and `git --staged` modes.
- **squawk** — a probe migration adding a plain `UNIQUE` to `public.users` is reported
  through the git-mode ratchet; the same two findings land on 0119, which added one. An
  unresolvable merge base fails rather than reporting an empty list.
- **sqlc** — an added query is reported; the tree is clean once reverted. No drift today.
- **globs** — all four combinations fire now (web file and direct-child primitive; tokens,
  adoption and docs), checked one at a time with `lefthook run pre-commit --file`.
- The `squawk` binary is resolved from `node_modules` rather than PATH: the hook runs the
  script directly with no `pnpm run` wrapper, so a bare `squawk` would have failed for
  everyone — on the one commit that actually carried a migration. Reproduced with a
  stripped PATH before the fix.

## Not in this PR

`knip` (dead JS code and unused dependencies) is the remaining gap. It is a different class
of check — hygiene rather than correctness or safety — and needs its own SvelteKit entry
configuration, so it belongs in its own change.

## Summary by CodeRabbit

* **Quality & Validation**
  * Added automated checks to keep generated database code synchronized.
  * Added linting for newly added SQL migrations.
  * Strengthened secret scanning for commits and complete repository history.
  * Added verification safeguards for security-scanning tools.
* **Developer Experience**
  * Added pre-commit checks for database code, migrations, and exposed secrets.
  * Documented migration, database-generation, and secret-scanning workflows.
  * Improved design-system check triggers and execution.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,6 +104,44 @@ jobs:
           path: coverage/*.out
           if-no-files-found: ignore
 
+  # Lints only the migrations this pull request ADDS. The applied history carries 1322
+  # findings and is never going to be fixed — "never edit an applied migration" is the
+  # rule this repository runs on, and rewriting one is a worse hazard than any finding.
+  # A migration file genuinely never changes after it lands, which makes this a far
+  # cleaner ratchet than the line-based one the Go lint uses: no baseline file to drift.
+  #
+  # Read .squawk.toml for why the invocation is split in two passes.
+  migrations:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+        with:
+          # The script asks git which migrations are new since the merge base, so the
+          # base commit has to be reachable. A shallow clone would make every migration
+          # look either new or absent, and neither is a useful answer — which is why the
+          # script fails on an unresolvable base rather than reporting an empty list.
+          fetch-depth: 0
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@v6.0.10
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v7
+        with:
+          node-version: 22
+          cache: pnpm
+          cache-dependency-path: pnpm-lock.yaml
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      # The step prints how many files it examined. On a branch with no migration that
+      # is zero, which is the honest answer — and visible, rather than a green tick over
+      # a check that found nothing to do.
+      - name: Lint new migrations
+        run: pnpm check:sql
+
   design-system:
     runs-on: ubuntu-latest
     defaults:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,6 +104,37 @@ jobs:
           path: coverage/*.out
           if-no-files-found: ignore
 
+  # internal/platform/db is generated and committed. Nothing compared the two until now,
+  # so a query edited without `make sqlc` shipped the old Go and every job stayed green —
+  # the same class of gap the design-system `check:dist` step exists for.
+  #
+  # `make sqlc` rather than a pinned action, because the Makefile already holds the
+  # version (SQLC_VERSION) and it is the one developers run. Two pins would be two
+  # answers to the same question, and the drift between them would look exactly like
+  # stale generated code.
+  sqlc:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Regenerate
+        run: make sqlc
+
+      # The bare `git status --porcelain` is correct HERE and would be wrong in the
+      # pre-commit hook: a fresh clone has an empty index, so "differs from HEAD" and
+      # "is unstaged" have the same answer. See lefthook.yml for the other half.
+      - name: Assert the generated code matches the queries
+        run: |
+          if [ -n "$(git status --porcelain -- internal/platform/db)" ]; then
+            git status --porcelain -- internal/platform/db
+            git --no-pager diff -- internal/platform/db
+            echo
+            echo 'internal/platform/db is stale. Run `make sqlc` and commit the result.'
+            exit 1
+          fi
+          echo 'Generated code matches internal/platform/db/queries.'
+
   # Lints only the migrations this pull request ADDS. The applied history carries 1322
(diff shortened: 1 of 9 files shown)


strelov1:
Five of six taken, in 00e0142b. Not taken: sourcing `.gitleaks.toml` from the base commit instead of the checkout — it would mean a config change is never exercised by the PR that makes it, and the threat it addresses (a contributor editing the allowlist to smuggle a credential past review) is visible in the diff and is not the failure this scanner exists for, which is the accidental paste.

Two details in the canary were found by running it rather than by writing it: a GitHub PAT tail must be exactly 36 characters or the rule does not match, and the token has to be assembled by `printf` so the workflow file is not itself a finding when the next step scans the history.

PR #2242, opened 2026-08-31
Lint the workflows, check the docs' links, and scan the lockfile this series added

strelov1:
Three more things nothing was reading, and one gap this series introduced.

## The workflows themselves

Nothing reads a workflow file except GitHub, at the moment it runs one. An unknown runner
label, a mistyped `needs:` or a broken expression is discovered by a job failing — and for a
workflow that fires only on a tag or a schedule, discovered late. `actionlint` reads all
eight without running any, and picks up a new one on its own because the job passes no path
argument.

The half that found real things is the shellcheck it runs over every `run:` block. **Three,
all fixed rather than suppressed:**

- The Storybook assertion built a variable from `find` and passed it **unquoted** to grep,
  relying on word splitting to expand several paths — which a filename with a space turns
  into a green check over the wrong files.
- The extension's store-upload step parsed `ls` for the artifact path. Same fragility, on
  the step that decides what gets published.
- The sqlc drift message put `` `make sqlc` `` in backticks inside single quotes. Harmless
  as written, and one editing slip from the shell running it.

`go install` at a pinned version rather than a downloaded release: the Go module proxy
verifies the checksum against the transparency log itself, so the pin is the whole
supply-chain story — no second digest to keep in step.

## The lockfile this series added

Mine. The repository root grew a `pnpm-lock.yaml` when the repo-wide tooling (knip, squawk)
landed, and Snyk covered nothing at the root: the job named the three package directories,
and its `paths` filter meant a change to the root lockfile **did not even start the job**.
Dev dependencies, but they execute in CI and on every laptop — the same trust boundary as
anything else here.

A plain `snyk test` is right for it: the root manifest has no `workspaces` key, so the
`--all-projects` the other two need would reach into their directories and bill them twice.

## The documentation's links

This repository navigates by document. `AGENTS.md` is a table of links into the eleven
blocks, each block links into its packages, the architecture doc links across all of it — a
reader, human or otherwise, is meant to follow those rather than search. Nothing has ever
read them, so a rename breaks one silently, and a broken link is worse than a missing one:
it sends the reader somewhere confidently.

**Two were already broken, both in an ACTIVE change rather than an archived one.**
`centralize-lifecycle-notifications` opens by pointing at a narrative design under
`docs/superpowers/specs/`, in its proposal and again in its design — and that file has never
existed, in any commit. The design document those two point *away from* is itself the
design, all 168 lines. Fixed by saying so.

Two implementation notes that are the difference between a signal and a list:

- **Fenced blocks and inline code are blanked out.** Six of the ten findings on the first
  run were Go generics: `Run[C](ctx context.Context, claimer Claimer[C])` is a bracket
  followed by a parenthesis, which is exactly the shape of a Markdown link. Code is replaced
  with spaces rather than deleted, so reported line numbers still point at the right line.
- **`git ls-files`, not a tree walk.** A walk would re-derive `.gitignore` and read whatever
  a local build left behind; `node_modules` alone carries thousands of READMEs.

The hook carries **no glob**, though this is a check about Markdown: what breaks a link is
renaming its *target*, usually a `.go` or `.svelte` file, so a glob on `*.md` would miss
precisely the commit that does the damage. 0.26 s over the whole repository.

## Verification

- **actionlint** — reports a planted unknown runner label; clean over the tree as it stands.
- **doc links** — a planted link to a nonexistent file is reported with its line; 257 links
  resolve without it.
- Both new CI jobs need no package install, so neither adds meaningful wall-clock.

## Summary by CodeRabbit

- **New Features**
  - Added automated validation for relative Markdown links, including pre-commit checks and CI reporting with file and line details.
  - Expanded security scanning to cover root-level JavaScript dependencies and lockfile changes.

- **Bug Fixes**
  - Improved extension package discovery during publishing.
  - Strengthened workflow linting and Storybook stylesheet checks to reduce false failures.

- **Documentation**
  - Clarified development guidance for workflow linting and documentation-link validation.
  - Corrected design-document references.


strelov1:
Both taken. The indented-fence one is not hypothetical here — `docs/superpowers/specs` and `extension/docs` both nest code blocks inside numbered lists, so the column-zero pattern was reading their contents as prose.

On the second: resolving against git's index rather than the filesystem also fixes a split I had not noticed — `existsSync` says yes to anything a local build left behind, so a link into `node_modules` would have passed on the machine that wrote it and failed in CI, which is the one place nobody is watching. Directory prefixes are derived from the index too, since git tracks files and a link to a folder should still resolve.

Not taken: regression cases as a test file. The probes are in the commit message rather than in `*.test.mjs` because this repository has no test runner wired at the root — adding one for a 100-line script is more machinery than the script is.

PR #2244, opened 2026-08-31
Lint the three things this repository hands to someone else

strelov1:
The three artefacts this repository hands to someone else, none of which any job read: a
shell script the site serves to be piped into `sh`, the images that get deployed, and the
API description integrations are built against. A mistake in any of them is discovered by
whoever is on the receiving end.

## Shell scripts

`web/static/install.sh` is documented as `curl -fsSL https://freehire.me/install.sh | sh` —
a mistake in it runs **on a stranger's machine, as them**. `web/docker-entrypoint.sh` is
what starts the site.

Both are clean today, so shellcheck goes in without a backlog. Two details:

- The list comes from `git ls-files`, not from the job, so a script added next year is
  covered the day it is committed rather than the day someone remembers this job exists.
- An **empty list fails**. A check with no subject is not a passing check.

## Dockerfiles

Two rules are ignored in `.hadolint.yaml`, and the reason is not taste: pinning an exact
package version in a distro image trades one failure for a worse one. The version
disappears on the next security update, and the build then breaks weeks later, on an
unrelated commit, in whichever image happened to rebuild first. The base image tag is the
pin that actually reproduces a build.

**`DL3064` on `web/Dockerfile` is right about the class and wrong about those lines**, so
its argument is inline rather than in the config. A credential in `ARG`/`ENV` is recorded
in that stage's layer history and leaks with the image — but this is a build stage whose
only output is the built app, and the stage carrying `SENTRY_AUTH_TOKEN` is discarded,
never tagged and never pushed. That is true of three lines, not of the repository, which
is why it is not a config entry.

`info`-level findings do not fail: consecutive `RUN`s and a non-numeric user id are style,
and a Dockerfile that reads clearly beats one concatenated into a single layer.

## OpenAPI

`web/static/openapi.yaml` is 1791 lines of public contract, served as a static asset, and
nothing opens it. It is validated against the **specification**, not against Redocly's
style opinions: the question is "can a client generator read this", and the answer stops
being yes on a mistyped `$ref` or a schema that does not parse.

It validates clean at `minimal`. The nine warnings the `recommended` ruleset adds are
style, and gating on them would be a different decision than the one this makes.

## Verification

- **shellcheck** — both scripts clean; the job fails if the tracked-script list is ever
  empty.
- **hadolint** — a probe Dockerfile with an untagged `FROM` and a `RUN cd` fails the gate;
  both real ones pass with only info-level findings.
- **OpenAPI** — validates clean at the pinned Redocly version.
- Versions pinned to ones that exist and were run locally: hadolint `v2.15.1`, Redocly
  `2.49.0`. Both of my first guesses did not exist, which is its own argument for checking
  a pin rather than writing one.

## Summary by CodeRabbit

* **Quality Improvements**
  * Added automated validation for shell scripts, Dockerfiles, and the OpenAPI specification.
  * CI now reports relevant linting and specification issues before changes are merged.
  * Added consistent validation rules and documented intentional Dockerfile exceptions.

* **Documentation**
  * Updated project guidance to explain the new validation checks and their coverage.


strelov1:
All three taken.

The hadolint one is the better shape and I should have written it that way: a global ignore meant the next Dockerfile added here would silently inherit the exemption, including one installing something where the version does matter. The three lines carry `# hadolint ignore=` markers now — same rule this repo already follows for squawk suppressions and for the DL3064 finding on the Sentry build args. Verified that it does not travel: a fresh `FROM alpine` + `apk add --no-cache curl` now fails the gate while both real Dockerfiles pass.

`permissions` went in at the workflow level rather than on the one job — CI was the only workflow here without it, and a floor covers the jobs added later too.

PR #2319, opened 2026-09-02
Recognise the six ATS platforms crawled since today

strelov1:
Six source adapters shipped with live board files — `paycor`, `dayforce`, `gusto`,
`hiringthing`, `ukgready` and `hrmdirect`, 10,891 boards between them — and none of their
domains was in `internal/ingest/atsboard`. A pasted link on any of them resolved to nothing,
so link import could not read it and the crowdsourced board-contribution flow could not
onboard it. Every one of the six PRs named this as the follow-up.

## Three needed a new extraction mode

None of the existing modes says where these platforms put the tenant, so each got a mode in
the table's own idiom rather than an inline special case:

| mode | platform | board |
|---|---|---|
| `pathlocalepair` | dayforce | the first **two** path segments after an optional leading culture — `jobs.dayforcehcm.com/en-US/dcrusa/join-us` → `dcrusa/join-us` |
| `query` | paycor | a named query parameter — the path is identical for every employer, so only `?clientId=<board>` names the board |
| `hostcareers` | ukgready | `<host>/<tenant>` read from `<host>/ta/<tenant>.careers` |

- **dayforce**: one tenant runs several career sites and each is its own board, so the tenant
  alone is not a shorter board id but one the crawl rejects. The culture stays **out**: a
  posting keeps the same id in every culture it is translated into, which is exactly why the
  ingest adapter folds it off the board too (`sources.boardIdentity`). The canonical collapses
  to the culture-free site root — verified live, `jobs.dayforcehcm.com/dcrusa/join-us` answers
  200 with and without a culture segment.
- **paycor**: `queryBoards` holds the parameter name and the listing path per host, so the
  canonical collapses a posting and the career home onto one board.
- **ukgready**: the host selects the environment a tenant is hosted in and is not derivable
  from the tenant id, so the board carries both — the same self-describing shape `hostpath`
  and `hosttenantboard` already use. The host in a pasted link is by construction one that
  answers for that tenant, which is what makes it safe to keep.

The other three fit existing modes: **gusto** is `path` with `boards` reserved, **hiringthing**
is `host` across the 25 white-label domains its board file is actually on, and **hrmdirect** is
`subdomain`.

## What each rule declines

A false board is the expensive direction — the contribution flow records and pays for it — so
each platform's known board-less shapes are declined outright:

- `apply.hrmdirect.com` serves every tenant's application form, and read as a subdomain it is
  a company called "apply". The guard is scoped to that apex (`platformLabelsByApex`) and
  **not** global — `apply.recruitee.com` is a board `recruitee.yml` actually tracks, so
  declining the label everywhere would have dropped a board the catalogue already crawls.
  Review caught this; there is now a test for it.
- A gusto `/postings/…` URL ends in the **posting's** uuid and carries the job's title slug;
  the board (`<company-slug>-<company-uuid>`) appears nowhere in it, so the URL is declined
  rather than turned into a board no crawl can resolve. The only link back is the breadcrumb
  the posting page renders, which is `boardresolve`'s job, not a URL's.
- Dayforce's own machinery shares the tenants' host: `/api/…` is the listing API and `/_next/…`
  the app bundle, both linked from every page on the platform, and both would otherwise read
  as the career sites `api/geo` and `_next/static`. `boardresolve` takes the *first* recognised
  ATS URL in a fetched page, so it would meet them before any real one. The check runs on the
  segment the board would **start** at, not the URL's first — the platform's own paths take a
  culture prefix too, so `/en-US/api/geo/…` would otherwise walk straight past it.

## Cross-checked against the board files

`TestRecognizeReturnsTheBoardTheBoardFileStores` pins the extraction against boards copied
**verbatim** out of each shipped `sources/<provider>.yml`, one per board-id shape those files
contain — including ukgready's bare-apex host (`saashr.com/6199997`), its alphanumeric tenant
ids (`secure3.yourpayrollhr.com/ihr0568`) and each of its six host families, and four of
hiringthing's reseller domains. A rule that disagrees with the board file is worse than no
rule: it files an employer under a board no adapter can crawl while looking exactly like a
successful recognition.

On top of the sampled table test, every board in all six files was replayed through
`Recognize` while reviewing this. **10,837 of 10,891 agree exactly** — paycor 2442/2442,
gusto 790/790, hiringthing 1457/1457, hrmdirect 890/890 — and the 54 that do not are two
known, documented shapes:

- **53 dayforce sites** (of 3,082) publish in no English culture and are entered in
  `dayforce.yml` **with** their culture, so this recogniser returns a prefix of the stored
  board and a link to one of them reads as a new board. Keeping the culture instead would
  break the other 3,029, whose links carry `en-US` and whose stored board has no culture
  segment at all. Called out in a test comment beside the case that pins it.
- **1 ukgready board** is stored with a `www.` host, which `hostname()` strips package-wide.
  The bare host is crawlable and `boardIdentity` folds the pair onto the tenant, so the cost is
  one duplicate contribution — cheaper than teaching one mode to keep a label the rest of the
  package normalises away. Noted beside the rows.

The replay was a throwaway; it is not committed, because it would need a hard-coded allowlist
for those two the moment either file grows, and the verbatim samples already carry the intent.

## paycom stays out

Not an oversight. `internal/ingest/atsdetect` owns that shape as one of six the shared table
**deliberately** excludes, and `TestLocalShapesStayOutsideTheSharedTable` pins the split in
both directions. The reason is in that test: `atsboard` is the accept-set the contribution flow
**pays** from, so widening it to a platform is a money-affecting decision that needs its own
proposal — not a tidy-up folded into a follow-up to six other changes. The harvest path, which
is what `atsdetect` serves, already reaches paycom and validates every candidate against the
platform before committing it.

## Notes

- No `sources/*.yml`, no source adapter, no `registry.go`, no generated contracts touched —
  the change is three files inside `internal/ingest/atsboard`.
- `query` is the one mode whose row is not self-contained (it needs a second row in
  `queryBoards`), so `TestQueryModeRowsAreConfigured` fails if a future one is added without
  it — otherwise the host would match and then decline every link on it, looking supported and
  recognising nothing.
- One drive-by: a doc comment for `firstSegmentAfterLocale` had come to sit above
  `ukgTenantBoard`, which this change moved back onto the function it describes because it
  refactored that function.
- `gofmt -l .` silent; `go vet ./...`, `go test ./...` and `go vet -tags=integration ./...`
  all clean.

## Summary by CodeRabbit

* **New Features**
  * Added recognition support for additional applicant tracking platforms, including Gusto, Dayforce, Paycor, HRM Direct, UKG Ready, and HiringThing.
  * Improved board identification across localized URLs, query parameters, tenant-based paths, and reseller domains.
  * Added support for platform-specific labels and board configurations.

* **Documentation**
  * Documented the new extraction modes, supported platforms, and white-labeled ATS handling.

### DIFF
--- a/internal/ingest/atsboard/AGENTS.md
+++ b/internal/ingest/atsboard/AGENTS.md
@@ -19,15 +19,22 @@ nothing but the URL. No network, no database.
   exist, while one adapter serves both domains as `factorial`.
 - **Adding an ATS is one row plus one test case.** Extraction modes say where the tenant sits:
   `path` (first path segment), `pathlocale` (same, skipping a leading `xx-XX` locale),
-  `pathportal` (the segment before the posting, for SmartRecruiters' portal URLs — plus its
+  `pathlocalepair` (the first TWO segments after that locale — Dayforce's board is
+  `<tenant>/<site>` and one tenant runs several sites), `pathportal` (the segment before the
+  posting, for SmartRecruiters' portal URLs — plus its
   one-click apply form, `/oneclick-ui/company/<board>/publication/<uuid>`, which names the
   employer mid-path and would otherwise be read as a board called `oneclick-ui`),
   `pathnumeric` (like `path`, but the segment must be an all-digit id — PageUp's board is a
   numeric institution id, so localisation/section segments like `/cw/en/search` are not
-  boards), `subdomain`
+  boards), `query` (a named query parameter, because Paycor serves every board from one path
+  under `?clientId=<board>`; `queryBoards` names the parameter, and it is the one mode that
+  needs a second row — a test fails if it is missing), `subdomain`
   (leftmost DNS label), `subdomainchain` (every label under the apex, for a tenant nested under a
   regional instance like `<tenant>.global.huntflow.io`), `host` (the whole careers host IS the
-  tenant), `hostpath` (host + first path segment, for Workday).
+  tenant), `hostpath` (host + first path segment, for Workday), `hostcareers`
+  (`<host>/<tenant>` from `<host>/ta/<tenant>.careers`, for UKG Ready, whose host selects the
+  environment its tenant lives in), `hosttenantboard` (`<host>/<tenant>/<guid>`, for UKG Pro
+  Recruiting — a different product from UKG Ready, on different hosts).
 - **The mode must match how the ingest adapter addresses the board**, not how the URL reads. The
   board string is copied verbatim into `sources/<provider>.yml` and into the `external_id`
   namespace, so a truncated one is a board that 404s every crawl: Huntflow's adapter fetches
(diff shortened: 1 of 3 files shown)


strelov1:
Applied: `clientId` must now match `^[0-9a-f]{32}$` before it is taken as a board — the same shape guard `pathnumeric` carries for PageUp, and the shape all 2,442 boards in `paycor.yml` have. `?clientId=not-a-board` and a truncated key are now declined, with test cases for both, and `queryBoards` makes the pattern required rather than optional so the configuration test catches a future row that omits it.

PR #2331, opened 2026-09-02
Correct the record: pinning the protocol did not fix the fetch

strelov1:
The comment shipped in #2325 says protocol v0 fixes the release fetch. **It does
not**, and the measurement behind it was invalid by construction.

## Why "v0 10/10, v2 2/10" lied

v0 sends the `git-upload-pack` POST **only when the fetch has work to do**.
Every probe in that measurement ran against an already-current checkout, where
the whole exchange is a single GET that cannot fail. So the comparison put a
real code path against one that was never exercised.

Re-measured on a checkout genuinely behind main:

```
GET  /info/refs?service=git-upload-pack  -> 200
POST /git-upload-pack                    -> 401   ← v0 too
fatal: could not read Username for 'https://github.com': terminal prompts disabled
```

## What is actually known

- GitHub 401s the POST half of a fetch from this host at **any** protocol
  version. The repository is public and `curl` as the same user gets a 200, so
  no credential is missing from the box.
- It is **intermittent** — the 14:49 release that day went through; the next one
  did not.
- "No such device or address" is a **missing terminal**, not a refused
  credential. A scripted release has nowhere to prompt.

## What was tried and disproved

Both are now written into the comment, because both convinced at the time:

1. **A retry loop** — ten attempts, ten 401s. Reverted.
2. **`protocol.version=0`** — shipped on the invalid measurement above.

The pin stays because it costs nothing, but the comment no longer claims it is a
fix. `GIT_TERMINAL_PROMPT=0` stays because it genuinely earns its place: it
turns a hang on an unanswerable prompt into an immediate, legible failure.

## The real fix, and the way through meanwhile

The limit that bites is on **unauthenticated** operations, and this host crawls
hard enough to sit under it. Authenticating the remote is the durable fix — it
needs a credential provisioned on the box, which is an owner's decision rather
than a script's.

Until then the comment carries the manual route, which works because a bundle is
a file and the fetch that reads it never talks to GitHub:

```
git bundle create /tmp/hire.bundle origin/main
scp /tmp/hire.bundle root@host:/tmp/
sudo -u freehire git fetch /tmp/hire.bundle main:refs/remotes/origin/main
sudo -u freehire git merge --ff-only origin/main
```

This is the third diagnosis of the same failure from the wrong measurement in
one afternoon. The comment now records what was disproved, not only what is
believed — that is the part with any value left in it.

shellcheck clean.

### DIFF
--- a/deploy/bin/release.sh
+++ b/deploy/bin/release.sh
@@ -41,30 +41,46 @@ if ! grep -q "127\.0\.0\.1:${aport};" "$newconf" || ! grep -q "127\.0\.0\.1:${wp
   exit 1
 fi
 cd "/opt/freehire/src/${dir}-${new}"
-# protocol.version=0, and the reason is worth the paragraph because the error message
-# points at the wrong thing entirely. On 2026-09-02 every release began dying with:
+# On 2026-09-02 releases started dying here with a message that names the wrong thing:
 #
 #   fatal: could not read Username for 'https://github.com': No such device or address
 #
-# The repository is public, `curl` as the same user fetched the same URL with a 200, and
-# there is no credential anywhere on the box — yet git was asking to log in. Tracing the
-# exchange showed why. Git's protocol v2 makes TWO requests, and GitHub answers them
-# differently for this host:
+# What is established. GitHub answers the two halves of a fetch differently for this
+# host, at ANY protocol version:
 #
 #   GET  /info/refs?service=git-upload-pack  -> 200
 #   POST /git-upload-pack                    -> 401  www-authenticate: Basic realm="GitHub"
 #
-# Protocol v0 needs only the first and succeeds: measured 10/10 where v2 managed 2/10.
-# The 2-in-10 is what made this look intermittent and cost an hour — it is not a flaky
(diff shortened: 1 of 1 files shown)


strelov1:
Closing: this landed, in a fuller form, in #2342 ("Fetch the release over SSH; the protocol pin lasted three hours").

`deploy/bin/release.sh` on main already carries the correction this PR was opened to make — that the v0 measurement was invalid and the pin did not fix the fetch:

> The pin bought about three hours. Later the same day the SAME 401 came back on v0, which is the answer to what the paragraph above left open: the throttle is not on a protocol version, it is on ANONYMOUS object fetches from this address.

with the same evidence, `ls-remote` (GET) → 200 against `fetch` (POST) → 401, 3/3.

It also carries what this PR does not: the durable fix. `origin` is `git@github.com:strelov1/freehire.git` in both colors, `~freehire/.ssh/config` points github.com at the `agent_deploy` key, and the comment names what to check next time (the key, not the protocol).

So merging this now would replace a record that explains the fix with one that stops at the diagnosis — which is also what the conflict with main is about.

PR #2352, opened 2026-09-03
Release main to production once CI is green on it, from a timer on the host

strelov1:
Deploying has been a person typing an ssh command. This makes it automatic, and the interesting decision is the **direction**: the host polls GitHub rather than GitHub pushing to the host.

## Why not a self-hosted runner

It is the obvious answer and it is the wrong one here. This repository is **public with 105 forks**, and a runner takes jobs from GitHub — a fork's pull request naming `runs-on: self-hosted` executes that fork's code on this machine. Secrets stay out of a fork's reach; the machine does not, and this machine is production.

An SSH key in GitHub secrets also works, and restricted with `command=` in `authorized_keys` it can only trigger a release. But it still puts a production credential somewhere the host does not control, and needs the inbound path to stay open.

Polling inverts both:

| | self-hosted runner | SSH key in Actions | this |
|---|---|---|---|
| GitHub can run arbitrary code on prod | **yes** | no | no |
| inbound path required | no | yes | no |
| production credential lives in GitHub | no | yes | **no** |
| what a leaked credential buys | — | trigger a release | draw a green tick |

## How it works

`freehire-autodeploy.timer`, every 10 minutes:

1. compare `origin/main` against the commit `hire-current` is serving
2. require **every** check run on that commit to have COMPLETED and passed
3. `release.sh`
4. commit status back to GitHub, and Telegram either way

It gates on **CI**, not on the merge: a squash merge pushes to main and the workflows start after it, so "main moved" would ship a commit whose tests are still running. Anything unfinished waits for the next tick — which is what makes ten minutes comfortable rather than slow.

Armed by `AUTODEPLOY=1` in `/opt/freehire/.env`. Without it the script logs what it *would* have done and exits 0, so the timer can be installed and watched before it is trusted — same opt-in shape as `site-alert.sh`'s channel.

## The Telegram message is load-bearing

There is no workflow run to be red here. Without the message this is one more thing that can fail unnoticed — and that is this project's recurring failure rather than a hypothetical: the `.env` split muted `remind`'s email for weeks, `INGEST_SLOTS` skipped 56% of runs, both with exit 0.

A commit that fails twice is left alone until main moves, so a broken build cannot rebuild itself every ten minutes. The count lives in `/var/lib/freehire/autodeploy.state`, which survives a reboot on purpose — a machine that reboots mid-incident must not come back and retry the same doomed build.

## release.sh takes a lock now

Until this timer existed, *"a person is running it"* was the only lock there was — nothing on this host holds one, and systemd's refusal to start a second `oneshot` instance only ever covered the timer path. A scheduled release starting while an operator is mid-flip is a collision **this change creates**, so it closes it. Exit 3 is a distinct code so `autodeploy.sh` reads a hand-run release as "not my turn" rather than as a failure worth alerting on.

## Verification

Run against the live host:

- the up-to-date path exits 0 quietly (`up to date at 67f0665b2`)
- the check-runs read parses the real API: 11 runs on the deployed commit, `pending=0`, `failed=[]`
- both scripts shellcheck-clean, `check:links` clean

Also fixed while writing it: `${gh_token:+-H "Authorization: ..."}` unquoted word-splits into four arguments and hands curl garbage. shellcheck does not flag it; it is an array now.

## Known, deliberately not guarded

A tick landing in the seconds between a push and its workflows being **created** would read a short list of check runs as a full one. The window is seconds against a ten-minute cadence, and a run that exists at all starts `queued`, which the pending check already waits on. The comment records that the fix would be a minimum commit age — not a list of required job names, which would drift the moment CI gains a job.

## Not in this PR

Installing it. The units and script have to be copied to the host (`deploy/` never deploys itself), and it needs `GITHUB_STATUS_TOKEN` — a fine-grained token with **Commit statuses: write** and **Checks: read** on this repository only.

## Summary by CodeRabbit

- **New Features**
  - Added automated production deployments that run every 10 minutes after startup.
  - Deployments proceed only when the latest commit passes all required checks.
  - Added safeguards to limit repeated failed attempts and provide deployment status alerts.
  - Added protection against overlapping releases to prevent conflicting deployments.

- **Documentation**
  - Updated deployment documentation with the automated deployment workflow, scheduling, safeguards, and current system inventory.

### DIFF
--- a/deploy/AGENTS.md
+++ b/deploy/AGENTS.md
@@ -9,8 +9,8 @@ Snapshot taken 2026-09-01 from `/etc/systemd/system/freehire-*` and `/opt/freehi
 ## What is here
 
 ```
-systemd/   337 files — 46 .service, 286 .timer, 5 drop-in directories
-bin/        13 operator scripts (release, backups, alerting, ingest slotting)
+systemd/   350 files — 48 .service, 297 .timer, 5 drop-in directories
+bin/        14 operator scripts (release, autodeploy, backups, alerting, ingest slotting)
 ```
 
 `systemd/freehire-ingest@.service` is one template; the 255 `freehire-ingest@<board>.timer`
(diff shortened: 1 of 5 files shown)


strelov1:
Wrong repository — host configuration belongs in freehire-ops. Redoing it there.

PR #2355, opened 2026-09-03
Filter jobs by how long they have been open, not only by the date the board states

strelov1:
## Why

A job carries two dates and the feed could only filter on one.

`posted_at` is what the board claims; `created_at` is when we first saw the posting. Some
boards rewrite the first on every crawl, so `?posted_within_days=3` returned postings
whose own card read `Open 72d` beside `37 minutes ago` — the filter and the badge
disagreeing in public. `jobreality` already detects that gap as `fake_freshness`, but the
knowledge only ever reached a badge.

## What this does

- **`open_within_days=N`** — a second bound, on `created_ts` (when ingest first wrote the
  row), which no source can rewrite. Same lenient parse as its neighbour, and the two AND,
  so "open within 30 days but posted within 3" is now expressible.
- **The `Posted` pane holds both bounds**, each labelled for whose date it reads, plus the
  `reality` facet — one pane is now the single answer to "how current is this posting".
- **The `Hide evergreen` toggle is gone.** It wrote one sign of one value of a facet, and
  it was the widest control in a toolbar that overflowed a 390px viewport. No parameter
  changed: excluding `Likely evergreen` in the pane writes the same
  `reality_exclude=likely-evergreen`.
- **The select above the list becomes `Open`** once the flag is on — that is the question
  its placement implies and the one the source date was quietly failing. With the flag off
  it bounds the source date exactly as before; there is never a state without a date
  control there.

No breaking changes. Every existing parameter keeps its spelling and its meaning.

Spec: `openspec/changes/split-posted-and-open-dates` (three delta specs — `job-search`,
`filter-modal`, `jobs-list-controls`; the last two currently *forbid* the pane change,
which is why both halves ride one branch).

## Deploy order — this part is load-bearing

Documented in `internal/search/search/AGENTS.md` ("Adding a filterable attribute"). Two
hazards, and both look like an outage rather than a missing feature:

1. A binary asking for an attribute the live index has not declared → Meili `400` → the
   handler maps it to **500 for every filtered search**.
2. Declaring the attribute does not retro-fill existing documents. The incremental drain
   only pushes documents whose `content_hash` moved, and `created_ts` is not in that hash
   — the trap `is_tech` and `requires_clearance` both fell into. A thin feed, not an
   error, so **nothing alerts**.

So, in order:

1. Patch the **live** index settings to declare `created_ts` filterable. Cheap; documents
   lag harmlessly.
2. Deploy this. The API accepts `open_within_days` from here; results are incomplete
   until step 4.
3. Confirm **≥45 GB free**, then `systemctl stop freehire-reindexw.timer`.
4. Full `make reindex` (`REINDEX_DEDUP` unset). Budget hours, not the ~26 min the older
   note quotes — a rebuild now also writes ~10 GB of skill vectors.
5. Verify by hand: `?open_within_days=3` against a posting known to carry a rewritten
   posting date.
6. `PUBLIC_OPEN_WITHIN=1`, restart web, re-enable the timer.

**Rollback:** unset the flag and restart web. Removing a filterable attribute needs no
dance — the binary stops asking first, and a stale declaration is harmless until the next
rebuild.

## Verified

- `go test ./...`, `go vet ./...`, `go vet -tags=integration ./...` — clean
- `pnpm vitest run` — 1319 passed (112 files); `pnpm check` — 0 errors
- OpenAPI document validates
- In a browser against the prod API, both flag states: on → `Open` select writing
  `open_within_days`, two labelled sliders, reality chips, no standalone
  `Posting reality` rail entry; off → select reads `Posted`, **and a shared link
  carrying the open bound still applies and still shows a removable chip**

## One thing worth a second look

`docs/API.md` is generated from `web/src/lib/docs` and was **already stale on main** — it
still described the retired AI-credits system and `GET /me/credits`, carried an outdated
category list, and omitted `requires_clearance`. Regenerating swept that in, so the diff
there is wider than this change. Trimming it to my own lines was not an option: the file's
header declares it and the rendered `/docs/api` page to be one source of truth, and a
hand-edited generated file breaks exactly that.

## Summary by CodeRabbit

- **New Features**
  - Added an optional “Open within” filter for jobs first recorded in the catalogue within a selected number of days.
  - “Open within” and “Posted within” filters can be applied independently or together.
  - Added corresponding filter controls, URL support, API documentation, and feature-flag configuration.

- **Updates**
  - Consolidated posting-reality options into the Posted filter pane.
  - Clarified date-filter descriptions and corrected API documentation links.

- **Removed**
  - Removed the standalone “Hide evergreen” toolbar control.


strelov1:
Addressed the review. Four of five findings were real; all four are in `b923d690`.

Three of them were **pre-existing on `posted_within_days`** — this change copied the
shape rather than introducing it — so the fixes went into the shared helpers and cover
both bounds.

| Finding | Verdict | What it actually was |
|---|---|---|
| Duration overflow | **confirmed, worse than reported** | `open_within_days=200000` — the most *permissive* input possible — returned an **empty list** (cutoff landed in 2063). Now bounded by `maxWithinDays` (a century); anything past it imposes no restriction, like a zero. |
| Parser grammars diverge | **confirmed** | `?open_within_days=1e2` drew an active "Last 100 days" chip over a list the server never bounded. `positiveDays` now matches `strconv.Atoi` exactly — including the leading zeros and leading `+` that Atoi *does* accept. |
| TOC anchor | **confirmed, 9 links not 1** | `slug()` collapsed a *run* of punctuation-and-space into one hyphen; renderers *drop* the punctuation and keep the surrounding spaces. Every heading with an `&` was dead, plus `Profile & résumé` — the old rule flattened the `é` too and produced `profile-r-sum`. All 39 resolve now. |
| Negative-only reindex check | **confirmed, best of the five** | A document still missing `created_ts` fails `created_ts >= …` exactly as a genuinely old one does, so "the stale posting is gone" passes equally well when the rebuild never ran. Step 7.5 now requires a recently-created posting to be **returned**, at catalogue-scale counts, before the flag goes on. |

**Skipped:** *"Close the test blocks"* in `filterSections.test.ts`. The suggested code is
character-for-character what is already on lines 109–114; the file parses and its 8 tests
pass. Reading it as truncated looks like a diff-hunk artefact.

`maximum: 36500` is now documented on both params in `openapi.yaml`.

PR #2361, opened 2026-09-03
An unbuilt suggestion dictionary is no suggestions, not a 500

strelov1:
Found by verifying #2358 on production, which is the only place it could be found.

Between the deploy and the first `cmd/build-suggestions` run the `suggestions`
index does not exist, so `GET /api/v1/suggest?q=…` answered **500**. The search
box asks for a completion on every settled keystroke, so that whole window — a
day, until the timer fires — was a broken-looking dropdown and a stream of errors
in the tracker, for a state that is simply "not built yet".

```
$ curl -s -o /dev/null -w '%{http_code}' 'https://freehire.me/api/v1/suggest?q=product+own'
500
```

Detected by HTTP status rather than by the engine's error code: the SDK keeps
that code in an unexported type, so no caller can construct one and a check
written against it could never be tested. A 404 from a search against this one
index has a single meaning anyway.

Every other failure stays a failure — an unreachable engine, a rejected filter, a
wrong key. Swallowing those would turn an outage into a box that quietly offers
nothing.

## Summary by CodeRabbit

* **New Features**
  * Added intelligent search-box suggestions for roles, categories, skills, companies, and job titles.
  * Suggestions support phrase completion, typo-tolerant matching, posting counts, keyboard navigation, and applying multiple filters at once.
  * Added curated starter suggestions when the search box is opened.
  * Search suggestions refresh automatically from current catalogue data and search demand.

* **Improvements**
  * Search text is now committed on Enter, selection, or clear, providing a more predictable typing experience.
  * Added dedicated API protection and graceful handling when suggestions are unavailable.


strelov1:
Superseded: the branch was already squash-merged, so it conflicts wholesale. Reopening the one commit off a fresh main.

PR #2405, opened 2026-09-04
Keep a skill chip's tooltip inside its own container, and slow its hover trigger

strelov1:
## Summary
- `Tooltip`: add a 300ms hover-intent delay before opening on mouseenter (focus and touch stay immediate), so gliding the pointer across a row of skill chips no longer flashes every tooltip it passes over.
- `Tooltip`: shift the floating content back inside an explicit `data-tooltip-boundary` ancestor when one is marked, falling back to the viewport otherwise — fixes a chip's tooltip spilling out of the job page's narrow sidebar card into the description column.
- `JobView`: move the skill chips out of the sidebar and into the end of the Description tab.
- `CountryFlagStack`: add a `labelClass` override so the job facet list's Country value matches its sibling facts (`font-medium`, ordinary foreground) instead of the component's default muted caption.

## Test plan
- [x] `pnpm test` in `design-system/` (151 passed, incl. new Tooltip delay/boundary coverage)
- [x] `pnpm test` in `web/` (1378 passed)
- [x] `pnpm run check` (svelte-check) in both packages — 0 errors
- [x] `pnpm run check:tokens` in `design-system/` — clean
- [x] `eslint`/`oxlint` on all changed files — no new issues
- [x] Verified live against real job data via `web/`'s dev server (`VITE_API_URL` pointed at the public API)

## Summary by CodeRabbit

* **New Features**
  * Tooltips now open after a brief hover delay, while focus and touch interactions remain immediate.
  * Tooltips automatically reposition to remain visible within the browser viewport.
  * Job skills now appear in the description under a clearly labeled section.
  * Country flag labels support customized styling for improved presentation.

* **Bug Fixes**
  * Prevented tooltips from opening when the pointer leaves before the hover delay completes.
  * Improved tooltip positioning when content is wider than the viewport.

### DIFF
--- a/design-system/src/tooltip.svelte
+++ b/design-system/src/tooltip.svelte
@@ -30,13 +30,36 @@
   const HIDE_DELAY_MS = 150;
   let hideTimer: ReturnType<typeof setTimeout> | undefined;
 
-  function show() {
+  // A row of chips fires mouseenter/mouseleave on each one a gliding pointer passes on
+  // its way to somewhere else. Opening with no delay meant every chip along the way
+  // flashed its tooltip open and immediately queued its own close — a strobe, and a
+  // burst of scheduled work for chips nobody meant to read. Requiring the pointer to
+  // sit still for a beat first ("hover intent") means a glide never arms the timer, so
+  // nothing but the chip actually being read opens. Only the delayed path
+  // (`scheduleShow`, wired to mouseenter) needs this — focus and a touch tap are each
+  // already one deliberate action, so `show` stays immediate for them.
+  const SHOW_DELAY_MS = 300;
+  let showTimer: ReturnType<typeof setTimeout> | undefined;
+
+  function clearTimers() {
+    clearTimeout(showTimer);
     clearTimeout(hideTimer);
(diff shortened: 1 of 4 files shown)


strelov1:
Ran `/code-review` on this branch. It found a real bug and one piece of now-dead code, both fixed in the follow-up commit:

1. **Double-edge overflow bug (fixed)** — `edgeShift()` only ever corrected one side. When the tooltip content is wider than the boundary itself (long text on a narrow phone viewport), both edges overflow at once, and the old logic pushed the near edge out even further while "fixing" the far one. Now prioritizes keeping the near (reading-start) edge visible, added a regression test for the double-overflow case.
2. **Unused `data-tooltip-boundary` marker (removed)** — after moving skills out of the sidebar into the Description tab, no call site ever set this attribute anymore. Kept it in place would have been speculative infrastructure with no consumer, so it's gone along with its test coverage; the viewport-only fallback is what every current caller actually relies on.
3. A dormant `side`-change reset was also tightened (not reachable by any current caller, but cheap to fix while in there).

PR #2414, opened 2026-09-04
Read the statuses the way the data actually uses them

strelov1:
Follow-up to #2412, found by the prod dry run before `--apply` ever ran.

## What the dry run said

```
404 contributions
  26  submission
   2  pendingBoard
 171  attribution
  14  refusal
 191  NEEDS A HUMAN: a recognized status carrying no board
```

Expected 26 / 2 / 172 / 204 and **zero** unplaceable. Two rules were wrong, and neither shape occurs in any fixture — only real data could show them.

## 1. A refusal is a refusal, classified or not

190 of the 204 refusals carry a **NULL source**. A refusal during triage leaves an unclassified URL exactly as it was — which is what migration 0037 made the column nullable for.

`destinationOf` checked for a board *before* reading the status, so all 190 came out "needs a human". The status now decides first, and only the two destinations that actually name a board go looking for one.

## 2. An empty board is not missing data

The 191st was an `onboarded` contribution on `amazon` — a **boardless** provider. It crawls one company's own API and has no board id; its catalog row holds `board = ''`.

Which providers are boardless is `boardcatalog.Validate`'s to know, so the check is gone rather than guessed at here.

## 3. The dry run exited 0

While reporting 191 unplaceable rows. That is the run whose whole job is to say whether the drop is safe yet, so it now carries the same nonzero exit the apply run does.

## Testing

Both shapes are in the table test, named after what prod showed. Unit + integration suites green, full build clean.

## Summary by CodeRabbit

* **New Features**
  * Added a one-time backfill worker to migrate legacy contribution records into the current board and submission system.
  * Supports dry-run reporting and an explicit apply mode, with idempotent processing and unresolved-record reporting.
  * Preserves original contribution timestamps and submitter attribution where applicable.

* **Documentation**
  * Documented the backfill workflow and required sequencing before removing legacy contribution data.

* **Tests**
  * Added coverage for status routing, timestamp preservation, attribution, error handling, and repeatable execution.


strelov1:
CI would not trigger on this branch (it carried pre-squash history). Reopened clean off main as a fresh PR.

PR #2443, opened 2026-09-04
Sell Pro in the app stores, without letting one provider revoke another's

strelov1:
Adds RevenueCat beside Stripe so Pro can be sold inside the mobile app, and reshapes `users.pro_until` so the two providers cannot erase each other.

Apple's guideline 3.1.1 and Google Play's Payments policy require digital content consumed in an app to be sold through in-app purchase, so the mobile client cannot use the Stripe checkout. The client half is a separate change in `freehire-mobile`; this is the contract it builds against.

## The column stops being written

`SyncUser` used to derive a timestamp from Stripe's view of a customer and write `users.pro_until` outright. Point a second provider at that and they revoke each other: an account that bought in the App Store has no Stripe subscription, so the next Stripe sync derives the zero time and takes away what Apple was paid for.

```
pro_until_stripe      the web subscription
pro_until_revenuecat  the App Store or Google Play subscription
pro_until_granted     given by hand, and by add-invites later

pro_until   GENERATED ALWAYS AS (GREATEST(...)) STORED
```

The point is not that `GREATEST` computes the right answer — it is that the wrong answer becomes **unwritable**. `UPDATE users SET pro_until = …` now fails with `428C9`. Every reader is untouched: `plan.TierOf`, `GetProUntil` and `/me/plan` read the same column with the same meaning.

Existing values are split by the only evidence available — an account holding a `stripe_customer_id` transacted with Stripe and belongs where a cancellation can revoke it; one without was hand-set and belongs where no provider will undo it.

## The seam is cut at account addressing

Not at "verify a signature and read state", which was the first sketch. Stripe stores a customer id and resolves an account two ways because a first purchase arrives before any binding exists; RevenueCat's `app_user_id` **is** `users.id`, so there is nothing to bind. Checkout, the portal, prices and receipts stay off the interface — they have no store counterpart, and four methods answering "not applicable" would be a union type wearing an abstraction's clothes.

The package splits into an unexported `engine` (true of both providers) and `Service` (Stripe, embedding it). `cmd/billing-sync` already called exactly the engine's methods.

## Three defects found while building it, all of them quiet

- **The unprocessed-events queue was not scoped by provider.** A Stripe pass handed a RevenueCat row would have written the wrong column and then *stamped the row processed* — a store purchase marked done having conferred nothing, never retried.
- **The near-expiry window predicated on the derived column.** It is the *furthest* of three sources, so a Stripe customer whose store subscription reaches beyond their renewal would never fall inside the window, and the lost renewal the pass exists to repair would stay lost.
- **The stranger check bound the read instead of the pass.** A first-time buyer has no recorded event and a NULL column — that is what "first" means — so every recovery path refused them at once, and a purchase whose webhook was lost could never confer anything. It now binds the bulk passes; `SyncCaller` is the authenticated self-service entry.

## Two places that fail closed on purpose

A null `expires_date` means non-expiring, and the column cannot say "forever". It says *"for as long as we keep confirming"*: a horizon the reconciler renews. The date fields are read as `json.RawMessage` so **absent** and **null** stay distinguishable — collapsed into pointers, a `{}` from a renamed field would have granted a month of Pro nobody bought.

RevenueCat's signature window is 90 minutes rather than Stripe's five, because their documentation does not say whether retries are re-signed and their last retry lands at 80 minutes. A too-narrow window drops paid subscriptions; a too-wide one lengthens the life of a replay that is already idempotent by event id. Task 7.2 says how to settle it.

## New surface

- `POST /api/v1/billing/revenuecat/webhook` — HMAC over the raw body. **No fallback to the shared `Authorization` header**, which is a replayable bearer credential.
- `POST /api/v1/billing/revenuecat/sync` — cookie-authenticated, names nobody but the caller, rate-limited. It closes the window between a purchase completing on the device and the webhook arriving.
- `GET /api/v1/me/plan` gains `pro_source`. Behavioural rather than informational: Apple forbids sending an in-app subscriber to a web page to cancel, and offering an in-app purchase to a Stripe subscriber sells them the same plan twice.

## Deploying

**Migration 0132 runs manually on prod BEFORE the deploy.** It rewrites the table, which at the 1397 rows `0129` measured is milliseconds — the ordering is what matters, not the duration. Run it through `./migrate`, which wraps a file in a transaction; `make migrate` and initdb use plain `psql -f`, and this file must not half-apply.

`deploy/rollback/` holds the reverse and the re-apply. The reverse **keeps** the three source columns: dropping them destroys which origin conferred each plan, and re-applying over that re-splits by `stripe_customer_id` and moves money in both directions. Both files are executed from disk by the integration tests, round trip included.

Granting Pro by hand is now `UPDATE users SET pro_until_granted = …`; the old statement fails on purpose. `deploy/AGENTS.md` carries that and the RevenueCat dashboard steps.

## Verification

`go build ./...`, `gofmt -l`, `go vet -tags=integration ./...`, `golangci-lint` and `pnpm check:sql` all clean. `go test -count=1 -tags=integration` passes across `identity/billing`, `api/handler`, `cmd/billing-sync`, `platform/db` and `ai/plan`.

**Not verified, and cannot be from a checkout:** no real purchase, no real webhook, no dashboard. Six tasks in `openspec/changes/add-store-purchases/tasks.md` are marked `NEEDS …` — store products, the RevenueCat project and webhook, the environment variables, and the prod migration.

## Summary by CodeRabbit

* **New Features**
  * Added RevenueCat support for App Store and Google Play purchases.
  * Added authenticated subscription sync and webhook processing for RevenueCat.
  * Pro plan responses now identify the active entitlement source: Stripe, RevenueCat, or a grant.
  * Pro access from multiple sources is preserved and combined automatically.
  * Billing reconciliation now processes Stripe and RevenueCat independently.

* **Bug Fixes**
  * One provider’s outage no longer prevents the other provider from being reconciled.

* **Documentation**
  * Added setup, API, migration, rollback, and manual Pro-grant guidance.


strelov1:
## Before merging: run 0135 on prod

`bin/autodeploy.sh` pulls `main` on a ten-minute timer, so merging **is** deploying. This branch adds three columns and turns `users.pro_until` into a generated one; the code written against them will start running within ten minutes of the merge. On the current prod schema that means `/me/plan` 500s for every caller, the Stripe webhook fails to apply, and `billing-sync` errors every hour.

So the order is not a preference:

```bash
# 1. On the host, look at the accounts the split cannot decide from evidence alone.
#    A Stripe customer whose expiry was hand-set by support lands in pro_until_stripe,
#    where the next sync shortens it. At ~1400 accounts this is a handful of rows.
psql "$DATABASE_URL" -c \
  "SELECT id, pro_until FROM users WHERE pro_until IS NOT NULL AND stripe_customer_id IS NOT NULL;"

# 2. Run the migration through the runner, which wraps a file in a transaction.
#    `make migrate` and initdb use plain `psql -f`, which does NOT — and this file
#    must not half-apply: between the DROP and the ADD there is no pro_until at all.
./migrate

# 3. Merge this PR. autodeploy takes it from there.

# 4. Afterwards, move any account from step 1 that was a hand-set grant:
#    UPDATE users SET pro_until_granted = pro_until_stripe, pro_until_stripe = NULL WHERE id = …;
```

Rolling back needs no improvisation: `deploy/rollback/0135_pro_until_sources.down.sql` restores a writable `pro_until` and **keeps** the three source columns, and `…reapply.sql` puts the generated column back without re-splitting anybody. Both are executed from disk by the integration tests, round trip included.

Granting Pro by hand changes with this: `UPDATE users SET pro_until = …` now fails with `428C9`, deliberately. Support grants go to `pro_until_granted`, which no provider sync can revoke.

`REVENUECAT_API_KEY` and `REVENUECAT_WEBHOOK_SECRET` in `/opt/freehire/.env` are what switch the store provider on. Until they are set, every RevenueCat route is unmounted and its reconciler pass is skipped — Stripe carries on untouched, so this can merge and deploy safely before the dashboard exists.

strelov1:
**Correction to the note above: the release applies this migration. Do not run it by hand.**

I read the fleet wrong. `deploy/bin/release.sh` builds `cmd/migrate` and runs it **before the new colour starts** — "so that colour never serves a request against an older schema", in its own words — and a migration failure aborts the release with the live colour untouched. `autodeploy.sh` calls `release.sh`. So merging applies `0135` and then deploys, in that order, under an advisory lock, idempotently.

The steps in the previous comment would have had somebody doing by hand, unlocked, what the pipeline already does properly.

**What is still worth doing before the merge**, because no pipeline can do it:

```sql
SELECT id, pro_until FROM users WHERE pro_until IS NOT NULL AND stripe_customer_id IS NOT NULL;
```

Those accounts land in `pro_until_stripe`, which is right unless support hand-set a longer expiry on top of a real subscription — the data cannot tell the two apart, and the next sync would shorten the second kind. At ~1400 accounts this is a handful of rows; move any that were grants to `pro_until_granted` afterwards.

**The one gap the release leaves.** Between the migration and the flip, the old binary briefly serves against the new schema, and its only write to this column — `SetProUntil` — fails with `428C9`. It fails *closed*: a Stripe delivery in that window goes unacknowledged and is retried, and `billing-sync` repeats a sync that changes nothing. That the column refuses assignment rather than accepting a value nobody meant is what makes the window survivable.

**Rollback** needs no improvisation: `deploy/rollback/0135_pro_until_sources.down.sql` restores a writable `pro_until` and keeps the three source columns; `…reapply.sql` puts the generated column back without re-splitting anybody. Both are executed from disk by the integration tests, round trip included.

**Granting Pro by hand changes:** `UPDATE users SET pro_until = …` now fails with `428C9`, deliberately. Support grants go to `pro_until_granted`, which no provider sync can revoke.

**RevenueCat stays off until configured.** Without `REVENUECAT_API_KEY` and `REVENUECAT_WEBHOOK_SECRET` in `/opt/freehire/.env`, every store route is unmounted and its reconciler pass is skipped. Stripe is untouched, so this merges and deploys safely before the dashboard exists.

PR #2480, opened 2026-09-05
Give the sign-in brand panel an olive fill

strelov1:
## Summary
- Swaps `AuthBrandPanel`'s `bg-foreground`/`text-background` pair for the design system's brand tokens (`--brand`, oats-green olive) so the left panel on `/signin` is olive instead of near-black.
- Turns the illustration/carousel card into a solid light (`brand-foreground`) surface with olive-tinted mock content, since the previous translucent-white treatment read as near-invisible on the olive background.

## Test plan
- [x] `pnpm exec eslint src/lib/components/AuthBrandPanel.svelte` — clean
- [x] `pnpm run check` (svelte-check) — 0 errors (pre-existing warnings elsewhere, unrelated)
- [x] `pnpm dev` + manual check of `/signin?mode=login` and `/signin` (register) in both light/dark theme classes

## Summary by CodeRabbit

* **New Features**
  * Moved the account-setup checklist from Tracking to the Profile page.
  * Checklist links now open the relevant Profile section, including during in-page navigation and browser back/forward navigation.
  * CV and search-alert steps continue linking to their existing destinations.

* **Style**
  * Updated the credential-form branding panel with an olive color scheme and refreshed hover states.

* **Tests**
  * Added coverage for profile checklist steps and their associated sections.


strelov1:
Automated \`/code-review\` flagged three WCAG contrast regressions in the first commit — the opacity fractions (bullets \`/70\`, footer/social links \`/50\`, tab dots \`/30\`/\`/50\`) were carried over unchanged from the old near-black/near-white panel, but \`brand\`/\`brand-foreground\` only has ~5.1:1 headroom, so those fractions landed as low as ~1.5:1. Fixed in the follow-up commit:

- Bullets, footer note, and social links now use full-opacity \`text-brand-foreground\` (no headroom left to dim), with hover feedback via underline instead of a color fade.
- Tab dots moved from \`bg-brand/30\`/\`/50\` to \`bg-brand/80\` (+ full \`bg-brand\` on hover), clearing the 3:1 non-text contrast minimum against the now-opaque illustration card.

Verified with \`pnpm exec eslint\` and \`pnpm run check\` (0 errors) plus a manual look at \`/signin?mode=login\`.

strelov1:
One more catch during a cleanup pass: dropping the color-based hover on the GitHub-star link (previous commit) left \`GithubStars\`' own base \`hover:text-accent-foreground\` unmasked — \`cn()\`'s \`twMerge\` only removes a base class when the passed \`className\` supplies a same-group replacement, and \`hover:underline\` isn't one. So hovering the star count would have flashed the text/icon to near-black against the olive panel. Fixed by re-adding an explicit \`hover:text-brand-foreground\` alongside the underline.

\`pnpm exec eslint\` and \`pnpm run check\` both clean (0 errors).

Issue #1759, opened 2026-08-11
harvest: adzuna — onboard orphan companies' own ATS boards


strelov1:
Snapshot (prod, 2026-08-11): **196,565 active jobs / 22,904 active companies** via `adzuna`.

Task:

1. Run `go run ./cmd/harvest-orphans -aggregators=adzuna` to list companies the catalogue holds only through `adzuna` (needs `DATABASE_URL`).
2. Live-validate candidate boards with `go run ./cmd/harvest-boards <provider> orphans.seed.json` and commit the ones that check out to `sources/<provider>.yml`.
3. Re-ingest, confirm the onboarded companies now carry their own first-party postings.
4. Decide `adzuna`'s fate: does what's left carry postings with no ATS equivalent anywhere — genuinely exclusive to this platform? Keep crawling it, narrowed to that slice. Otherwise, propose deprecating `adzuna` as a source (drop it from `sources/`, retire its cron).

Part of #1758.

strelov1:
Claiming this one — starting with adzuna since it's the largest aggregator by active jobs/companies.

strelov1:
Batch 1/2 done — PR #1819.

Ran \`cmd/harvest-orphans -aggregators=adzuna\` on prod: 18,726 companies known to us only through adzuna → 38,076 name-derived board candidates. Then \`cmd/harvest-boards\` live-validated the candidates against 10 completed providers:

| Provider | Live boards |
|---|---:|
| smartrecruiters | 198 |
| pinpoint | 104 |
| bamboohr | 78 |
| greenhouse | 27 |
| ashby | 21 |
| recruitee | 37 |
| lever | 4 |
| gem | 9 |
| deel | 2 |
| teamtailor | 0 |

**480 boards onboarded.** breezy, freshteam, personio, trakstar, workable still running (workable is rate-paced, slowest) — batch 2 to follow.

strelov1:
## Step 4, answered: Adzuna is kept and narrowed, not deprecated

Measured on prod and against the live API on 2026-09-06.

### Is the remainder exclusive?

Classifying every open Adzuna posting by whether its company is reachable another way:

| the company is… | companies | postings | of which `is_tech` |
|---|---:|---:|---:|
| already held through its own ATS | 8,915 | 142,229 | 87,464 |
| carried by another aggregator | 6,517 | 72,931 | 45,471 |
| **reachable only through Adzuna** | **20,350** | **81,730** | 38,603 |

[...]

PR #2536, opened 2026-09-06
Aim the Adzuna budget at postings we do not already hold

strelov1:
## Summary

Adzuna is the catalogue's largest single source — **301,008 open postings, 4.34% of the
6,940,260 open catalogue** — and the crawl filling it has never sent a sort order, so the
platform answers in *relevance* order, which is stable between runs. Forty pages of that,
hourly, re-read one fixed slice for ever. Measured live on 2026-09-06, `gb/it-jobs`:

| request | reported total | first result's `created` |
|---|---:|---|
| what the adapter sent | 52,942 | **2026-02-10** |
| the same request with `sort_by=date` | 52,941 | 2026-09-06T12:34 |

That budget bought ~23,500 posting-slots a day and wrote ~6,300 new postings — a **27%
yield** — at **~470 requests/day**.

`sort_by=date` at 240 requests/day halves the requests and roughly doubles the intake:

| | before | after |
|---|---:|---:|
| requests/day | ~470 | **240** |
| new postings written/day | ~6,300 | up to ~12,000 |
| vs the 2,500/month ceiling | 5.6× over | **2.9× over** |

### It does not bring us inside the terms, and an earlier revision of this PR wrongly said it did

Adzuna states **four** ceilings and they disagree: `25/min · 250/day · 1,000/week ·
2,500/month`. 250/day multiplies out to 7,500/month, three times the monthly figure, so the
tightest governs and the real ceiling is **~83 requests/day**. The first draft read the two
figures it had as one thing said twice and did the arithmetic against the convenient one.

240/day is a **deliberate overage**: 2,500 × 50 is 125,000 posting-slots a month, a hard
ceiling no arrangement of pages, boards or runs can raise, so full compliance would cap intake
at ~4,150/day — *below* the ~6,300 the wasteful crawl manages today, for the source carrying
20,350 companies the catalogue holds nowhere else. The budget test now reports every ceiling's
multiple rather than passing on the strength of the one it enforces, and requesting the higher
limits Adzuna's terms invite ("available upon request for publishing use cases") is a task on
the change.

## What changed

- **`internal/ingest/sources/adzuna.go`** — sends `sort_by=date` and `max_days_old=2`; page
  budget 40 → 15. `max_days_old` is *not* redundant with the ordering: `Fetch`'s loop ends on
  an empty page, and against a 53k-deep feed that never happens inside the budget, so a quiet
  board (`au` publishes ~176 between runs) would spend its full budget on postings the
  seen-set discards — drawn from the same ceiling a busy board needs.
- **`deploy/systemd/freehire-ingest@adzuna.timer`** — hourly → four runs a day.
  4 boards × 15 pages × 4 runs = **240/day**.
- **`cmd/liveness/main.go`** — `adzuna` joins `expireDespiteRegisteredPrefixes`. Once the
  crawl is date-ordered it stops being a liveness signal for an ageing posting, and a probe
  cannot take over: following 40 stored URLs, **~30 answered `403 Access Denied`** (bot
  protection on the `…/land/ad/<id>` tracking redirect, which classifies `uncertain` —
  correctly). Same classification `whatjobs` already has, and it keeps that family's second
  clock: a posting must ALSO have gone unseen past its source's sweep grace, so one the crawl
  is still listing is never age-closed.
- **A guard that family already needed.** Every member is credential-gated and this worker
  builds the registry from that same constructor, so a host without `ADZUNA_APP_ID` drops the
  source out of the age rule *and* into the URL probe — the lifecycle mechanism changing hands
  on an environment variable about crawling, silently, in the direction that leaves postings
  open for ever. A configured prefix matching no registered provider now refuses to run.
- **`openspec/changes/narrow-adzuna-crawl`** — proposal, design, tasks, and two spec deltas.
  `job-lifecycle`'s age-rule requirement currently *forbids* what `expireDespiteRegistered`
  has done for every whatjobs market since it landed (the string appears nowhere under
  `openspec/`). Code is right, spec was stale; rewritten to state the real rule. `adzuna-source`
  is new — the largest source in the catalogue had no spec at all.

## Why Adzuna is kept rather than deprecated

Issue #1759's step 4. Classifying every open Adzuna posting by whether its company is
reachable another way:

| the company is… | companies | postings |
|---|---:|---:|
| already held through its own ATS | 8,915 | 142,229 |
| carried by another aggregator | 6,517 | 72,931 |
| **reachable only through Adzuna** | **20,350** | **81,730** |

72% redundant at the company level — and dropping it would still cost 20,350 employers the
catalogue holds nowhere else. Attention agrees narrowly: 4.87% of bot-filtered `page_uniques`
for 4.34% of the catalogue.

## Expected effect, stated up front

**The Adzuna slice will fall from ~301k toward 150-180k over several weeks**, faster at first
— its own 14-day sweep reaches postings before the 45-day age rule does. Intake roughly
doubles over the same period, so the slice refills with postings that are current rather than
with February inventory. Recorded in the proposal so the drop is not mistaken for a broken
adapter.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `go vet -tags=integration ./...`, `go test ./...`
- [x] gofmt clean; pre-commit hooks pass (golangci-lint 0 issues, govulncheck clean)
- [x] 5 new tests: the request carries both parameters; a quiet board stops at the empty page
  instead of spending its budget; the page budget stays under the daily ceiling; `adzuna`
  resolves without over-reaching a similarly-named provider; the unseen window does not narrow
- [x] API behaviour verified live, not assumed: `sort_by=date` and `max_days_old` both work,
  a malformed window answers 500 (fails loudly rather than returning unfiltered results), and
  `results_per_page=100` answers 400 — confirming the 50 the constant already assumed
- [x] Timer verified with `systemd-analyze calendar --iterations=4` and `systemd-analyze
  verify`. Worth flagging: **`*:22:00/6` parses** — as every hour at :22 repeating on the
  *second* — so it would have shipped the old hourly cadence while reading in review like a
  change. Nothing in the build reads a `.timer`. The unit records both wrong forms.
- [ ] **After merge, the timer must be copied to the host** — `release.sh` flips the app and
  never touches a unit, so until then the live half is a 15-page adapter still running hourly
- [ ] Post-deploy: count a day's real requests (expect **below** 240 — the early exit should
  make quiet boards spend less; at or above means it is not firing) and confirm intake moved

Closes the step-4 question in #1759.

## Summary by CodeRabbit

* **Improvements**
  * Adzuna listings are now prioritized by publication date and limited to postings from the past two days.
  * Crawls use up to 50 results per page and stop sooner when no recent results are available.
  * Adzuna crawling now runs four times daily instead of hourly, keeping request usage within service limits.

* **Reliability**
  * Expiration configuration is now validated at startup, preventing unmatched provider settings from being silently ignored.

* **Documentation**
  * Added documentation describing Adzuna filtering, crawl limits, freshness handling, and job expiration behavior.

### DIFF
--- a/cmd/liveness/main.go
+++ b/cmd/liveness/main.go
@@ -129,7 +129,18 @@ const staleCutoff = sources.DefaultSweepGrace
 // other), all sharing the identical unprobeable-URL shape. Matching by prefix against the
 // live registry (see the derivation in run()) means a newly onboarded market is covered
 // automatically instead of silently falling through an enumerated list.
-var expireDespiteRegisteredPrefixes = []string{"whatjobs"}
+//
+// adzuna reaches the same classification by a different route, and the obvious objection —
+// "its url is an adzuna page, so just probe it" — was tested rather than assumed. Following
+// 40 stored urls on 2026-09-06: roughly three quarters answered 403 Access Denied, Adzuna's
+// bot protection on the `…/land/ad/<id>` tracking redirect we store (it carries the
+// attribution parameters that credit a click to this publisher, so it is not the employer's
+// page and never will be). A 403 classifies uncertain, correctly — it is the platform
+// declining to answer, not a verdict — so a probe can never close an adzuna posting, however
+// long it runs. The leak on the other side is new as of the date-ordered crawl: a bounded
+// crawl reading newest-first never revisits an ageing posting, so the sweep stops being
+// offered it. Both halves of the membership test therefore hold.
+var expireDespiteRegisteredPrefixes = []string{"whatjobs", "adzuna"}
 
 func main() {
 	worker.Main(run)
@@ -213,8 +224,15 @@ func run() int {
 	}
 
 	// Derived from the live registry rather than enumerated (see
-	// expireDespiteRegisteredPrefixes) — every whatjobs market this run, no drift guard
-	// needed, since membership follows atsProviders by construction.
+	// expireDespiteRegisteredPrefixes), so a newly onboarded whatjobs market is covered this
+	// run without being named anywhere. That construction is why the list holds prefixes,
+	// and it is a complete answer for a member being ADDED — but not for one being ABSENT,
+	// which is what a missing crawl credential produces. Guard that case explicitly.
+	if missing := unmatchedPrefixes(atsProviders, expireDespiteRegisteredPrefixes); len(missing) > 0 {
+		log.Printf("liveness: %q matches no registered ATS provider — refusing to run (a source that cannot be age-expired would silently fall to the URL probe instead)",
+			strings.Join(missing, ", "))
+		return 1
+	}
 	expireDespiteRegistered := matchingProviders(atsProviders, expireDespiteRegisteredPrefixes)
 
 	// Appended AFTER the guard above so the empty-ATS-registry safeguard still keys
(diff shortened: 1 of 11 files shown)


strelov1:
Thanks — four of five taken, and the monthly-cap one was a real error on my side, not a nit.

**Monthly cap (major) — taken, and it went further than the finding.** The terms state
*four* ceilings, not two: `25/min · 250/day · 1,000/week · 2,500/month`. The weekly one was
in neither my notes nor this finding. They disagree (250/day is 7,500/month), so the tightest
governs and the real ceiling is ~83/day.

I did **not** size the budget to it, and the reason is in `design.md`: 2,500 × 50 is 125,000
posting-slots a month, a hard ceiling, so full compliance caps intake at ~4,150/day — below
the ~6,300 today's wasteful crawl manages, for the source carrying 20,350 companies the
catalogue holds nowhere else. 240/day is a recorded overage (2.9× versus today's 5.6×), the
test now logs every ceiling's multiple instead of passing on the one it enforces, and
requesting the higher limits Adzuna's terms invite is a task. The claim is corrected in the
proposal, the design, the source spec, the PR body and the issue comment that had already
been published.

**`gen-ingest-timers.sh` (major) — taken, and it was the find of the review.** The generator
writes the unit on the host and `adzuna` reached the default hourly branch, so running it
would have silently restored 4× the budget. It now has its own arm beside `reed`'s, which
exists for the same reason (a per-hour quota its full crawl blows past).

**`Persistent=false` (major) — taken.** One note beyond the finding: the generator's loop
needed a per-provider reset, or the first provider to turn catch-up off would have turned it
off for every provider generated after it.

**Exact `max_days_old` assertion — taken.** `strings.Contains` accepting `max_days_old=20`
for `2` is the regression that would look fine and quietly re-admit the stale inventory this
change exists to stop fetching.

**Guard ordering — not taken.** Both guards refuse the run, so the requirement ("the worker
SHALL stop") holds either way. With an empty registry `no ATS providers registered` is the
more useful message, and the proposed diff duplicates the check rather than moving it.

PR #2565, opened 2026-09-07
Delete a board's board_health row when it is retired

strelov1:
## Summary
- A retired board is never crawled again, so its `board_health` row (cooldown/failure count) could never clear the normal way (a successful crawl) — left in place it showed the provider as permanently unhealthy on the public `/status` page forever.
- Adds `DeleteBoardHealth` (sqlc) and calls it from both retirement paths: `cmd/prune`'s `retireBoards` and `boardcatalog.QueriesRepository.Retire`, right after the retirement itself actually persists (a no-op retirement deletes nothing).

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...` and `go vet -tags=integration ./internal/ingest/boardcatalog/...`
- [x] `go test ./cmd/prune/... ./internal/ingest/boardcatalog/... ./internal/platform/db/...`

### DIFF
--- a/cmd/prune/boards_test.go
+++ b/cmd/prune/boards_test.go
@@ -18,6 +18,9 @@ type fakeCatalog struct {
 	listErr error
 	// retired records every (provider, board, region) the run retired, in order.
 	retired []db.RetireBoardParams
+	// healthDeleted records every (provider, board, region) whose board_health row the
+	// run deleted, in order.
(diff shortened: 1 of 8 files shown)


strelov1:
Superseded by #2567, which merged the identical board_health cleanup as part of a bundled PR — the changes are already on main. Closing to avoid a duplicate/conflicting merge.

PR #2574, opened 2026-09-07
Show unique-job and new-signup counts, relabel the raw job count everywhere

strelov1:
## Summary
- `/open`'s "open jobs" figure counts every open posting, including reposts across boards (never duplicate-suppressed), which read far higher than the deduplicated total `/jobs` itself shows (~4.6M vs ~2.08M). Relabeled that figure to **"postings tracked"** everywhere it's quoted (homepage, `/about`, both OG cards, `/open`) and added a second, deduplicated **"unique jobs to browse"** figure on `/open` that matches `/jobs`' own search total. Extended the `/open` FAQ to explain the difference.
- Added a **"new members per day"** bar chart next to the existing cumulative member-growth chart on `/open`: `ListUserGrowth` now returns each day's own (non-cumulative) signup count alongside the running total, computed live from `users.created_at` in the same query — no new rollup/backfill needed.

## Test plan
- [x] `go build ./... && go vet ./... && go test ./...`
- [x] `go test -tags=integration ./internal/api/handler/ -run TestUserGrowthEndpoint` (seeds registrations, checks both cumulative and new-per-day series)
- [x] `npx vitest run` (new `signupsChart.test.ts`, unchanged `activityChart.test.ts`)
- [x] `svelte-check` — 0 errors
- [x] `eslint` on all changed frontend files — clean
- [ ] Manual check of `/open`, `/`, `/about` and their OG cards after deploy

## Summary by CodeRabbit

- **New Features**
  - The Open page now distinguishes total tracked postings from unique jobs available to browse.
  - Added a “New members per day” bar chart with daily signup counts and tooltips.
  - Statistics APIs now provide unique open-job counts and daily non-cumulative signup totals.

- **Improvements**
  - Updated homepage and social-image labels to use “postings tracked.”
  - Clarified that posting totals include reposts across multiple boards.
  - Catalog statistics remain available when the unique-job count cannot be calculated.


strelov1:
Addressed the two review findings specific to this PR's new code (commit 64606565):
- **Zero-as-failure-sentinel** (main.go:140): fixed — `resolveUniqueOpenJobs` now keeps the previously published figure on a failed/missing Meilisearch read instead of overwriting it with 0.
- **Unbounded Meilisearch call**: fixed — wrapped in a 10s `context.WithTimeout`.

Not fixed here, flagged as out of scope:
- **Non-HTTPS `MEILI_URL`** and **cross-origin redirect leaking the API key**: both apply identically to all 12 other existing `search.NewClient(cfg.MeiliURL, cfg.MeiliKey)` call sites in this repo (including `cmd/server` itself), not just the one this PR adds. Fixing it only here would be inconsistent and incomplete — the real fix belongs in `search.NewClient` itself, as a separate, dedicated follow-up.

PR #2583, opened 2026-09-07
Add a browser-use cloud fallback for Ashby/Workable submissions chromedp can't fill

strelov1:
## Summary
- Ashby and Workable already fully resolve a `Plan` through the existing schema-only `Reconcile`, but chromedp has no fill/submit path for either platform, so a fully-answered attempt still parked.
- Adds a narrow browser-use.com cloud-agent executor that only ever types an already-resolved `Plan`'s exact values — it never decides an answer itself, preserving every existing invariant (sensitive-keyword gate, geography park, "never guess").
- Ships behind `AUTO_APPLY_BROWSERUSE_ENFORCE` (shadow-first, OFF by default) plus a per-run `maxCostUsd` cap and a per-process-invocation spend guard.
- Recruitee was found unreachable during implementation (no registered `applyform.Fetcher`) and stays excluded.
- A live spike backing this change measured browser-use following exact field instructions precisely on a real Greenhouse posting, correctly leaving unlisted fields untouched and reporting a native validation failure honestly rather than fabricating success.
- Second commit applies findings from `/code-review`: a run erroring after creation now maps to `StatusUnconfirmed` (not a plain retryable error) since it may already be interacting with the live form; renamed the spend guard from "daily" to accurately reflect it resets every process invocation; added an explicit prompt-injection-defense sentence since field labels come from the employer's own (untrusted) ATS.

## Test plan
- [x] `go build ./...`, `go vet ./...`, `go test ./...` clean
- [x] `go vet -tags=integration ./...` clean (touches no `newAssistantHandlers`-style constructor)
- [x] Unit tests for `buildTask`/`parseOutcome`/`browserUseEligible`/`runSpendGuard`, plus fixture-based `Client.Submit` tests covering: fully-resolved plan executes and confirms; unresolved plan never calls browser-use; shadow mode never calls browser-use; unconfigured client parks as before; a run erroring mid-flight maps to unconfirmed, not a retryable error
- [ ] Manual shadow-mode dry run against one real Ashby/Workable posting before ever flipping `AUTO_APPLY_BROWSERUSE_ENFORCE=1` in production

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Added an optional browser-assisted fallback for submitting fully resolved applications on Ashby and Workable.
  * Applications are submitted only when all required answers are already resolved; résumé uploads and unsupported providers remain unchanged.
  * Added explicit outcome handling for confirmed, unconfirmed, and parked submissions.
  * Added per-run and aggregate spending limits, with shadow mode available before enforcement.
  * Enabled configuration through the `AUTO_APPLY_BROWSERUSE_API_KEY` environment variable.

* **Documentation**
  * Documented eligibility, confirmation rules, configuration, and spending safeguards.

### DIFF
--- a/cmd/auto-apply/main.go
+++ b/cmd/auto-apply/main.go
@@ -21,6 +21,7 @@ import (
 	"github.com/strelov1/freehire/internal/ingest/screeninganswers"
 	"github.com/strelov1/freehire/internal/ingest/sources"
 	"github.com/strelov1/freehire/internal/platform/blobstore"
+	"github.com/strelov1/freehire/internal/platform/browseruse"
 	"github.com/strelov1/freehire/internal/platform/config"
 	"github.com/strelov1/freehire/internal/platform/db"
 	"github.com/strelov1/freehire/internal/platform/llm"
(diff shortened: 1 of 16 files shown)


MagMueller:
are résumé uploads the next coverage gap you want to solve with Browser Use?

PR #2599, opened 2026-09-07
sources: add Keka adapter

strelov1:
## Summary
- Adds a source adapter for Keka (`<board>.keka.com/careers`), an Indian HR/ATS SaaS. A `board_submissions` row (`minfy.keka.com/careers/jobdetails/146875`) turned out to be a live Keka board during a manual queue drain.
- The plain `/careers/` bootstrap page carries no jobs itself — it fetches a static content fragment keyed by the org's own GUID, which the bootstrap page's own `fetch()` call embeds in its URL. Once that GUID is known, one request to the platform's embedded-jobs widget API (the same one an org embeds on its own site: `/careers/api/embedjobs/default/active/<guid>`) returns every open posting fully detailed — description included — so no per-posting detail fetch is needed at all.
- Registered `fullBoardListing`: verified live against an 88-posting board (Minfy Technologies), no pagination.
- `jobType` maps to `full_time`/`part_time` by reading the platform's own widget script's rendering logic (`job.jobType == 2 ? "Full Time" : "Part time"`) rather than guessing from field names.
- No structured remote flag is exposed for this board's configured job fields (`jobListingSetting.jobFields` lists only location/experience/jobType/dateOfPosting), so `Remote`/`WorkMode` fall back to the location/title heuristic like any adapter without one.
- The org's real display name (`api/organization/default/careerportalinfo`) is a best-effort lookup: its failure falls back to the board's configured `Company`, it does not abort the crawl.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./internal/ingest/sources/...`
- [x] `go test ./internal/ingest/sources/...` — new `TestKeka*` cases cover: provider key, `fullBoardListing` marker + registry check, `All()` registration, a missing org-GUID and a listing-fetch failure each aborting the whole board, a full listing round trip (title/company/location/employment-type/description/posted-at across two postings with different `jobType`s), and the org-name lookup falling back to the configured company name on failure without aborting the crawl.
- [x] `go run ./cmd/gen-contracts` (new `keka` provider in `SOURCE_VALUES`)

## Summary by CodeRabbit

* **New Features**
  * Added support for importing job listings from Keka-powered career portals.
  * Keka listings include titles, locations, employment types, descriptions, posting dates, remote indicators, and direct application links.
  * Added automatic company-name detection with fallback to the configured company name.
  * Registered Keka as an available job board provider.


strelov1:
Update: while harvesting more real boards for this adapter, found Keka actually serves career portals from **two different frontend template generations** (\`jobListingFormat\` 1 vs 2) — the org-GUID scrape only covered format 2. Both templates' own frontend call the same simpler, GUID-less endpoint (\`/api/jobs/{portalName}/active\`), so the adapter now uses that directly instead. Verified live against one board of each format (minfy/format 2, ihl/format 1) with unchanged and newly-working results respectively. Pushed as a follow-up commit on this branch.

PR #2645, opened 2026-09-08
Open the Talent Network: a public anonymous catalogue, and one button to be in it

strelov1:
The Talent Network shipped with a working page nobody could reach: no entry in the
account navigation, no button on the profile. A feature with a backend, a frontend, and
in practice no members.

This adds the surface that makes joining worth it — one public, anonymised catalogue —
and collapses the three visibility modes into a single toggle.

## What a visitor sees

`/talent` — the catalogue: server-rendered, filterable, indexable. Every control is an
`<a href>` that navigates rather than a local-state toggle, so a narrowed catalogue is a
shareable link, works with the back button, and shows a crawler what it shows a person.

`/talent/<handle>` — one card, carrying `noindex`. A card is a person; a search engine's
cache of one would outlive their decision to leave.

## The one rule

> Everything published is a value a **dictionary** resolved, a number, or a date. Nothing
> a candidate typed reaches a public response.

That is stricter than masking the fields that identify somebody, and the difference is
why this does not reuse `resumeextract.Anonymous`. Anonymous replaces the current role's
`company` column and leaves the rest alone — right for a link handed to one person, wrong
for a page a crawler reads. The employer's name is usually sitting in the prose beside
the column that was masked: in the summary, in the highlights, in the title
(`Backend Engineer @ X`), in a project's name, in the institution line.

Masking named fields fails toward disclosure — the next field the extraction contract
grows is published by default. A dictionary whitelist fails the other way.

**The test is the feature.** One CV, fifteen runs, each hiding a distinctive employer
token in exactly one place, asserting it appears nowhere in the marshalled card.

What the rule costs, recorded rather than quietly dropped: languages, certifications and
education are absent. All three are free text, and the only text→level dictionary lives
in `internal/job/jobfacts` — layer 5, which `candidate` may not import.

## Membership, and the handle

One toggle. The mode question — how much of yourself to disclose — is one the product
should answer, and it has.

A member is addressed by a **minted handle** (`backend-7f2a`), not by the account's
`username`. `internal/identity/username`'s own doc comment anticipates that use, but
`Suggest` derives the username from the email's local part, so for most accounts it *is*
the person's name — and the hosted mailbox adopts the same string as a live address.

The handle is frozen at mint: the seniority is deliberately left out of it, because a
grade is the part of a title most likely to change and a URL that moves is one somebody
has already shared.

## Two filters over one vocabulary

`specializations` is what a candidate ticked on their own profile — where they want to
go. `category` is derived from their most recent job title — where they have been. The
listing filters on the first; the second is on the card. Two rows over one 48-value
vocabulary would be 96 chips on a screen.

## Deploy order — please read

Three of the four migrations are ordinary: apply **before** the deploy.

`0151_drop_talent_network_public_id.sql` is the exception and says so at length in its
own header: **apply it AFTER the deploy.** The running server still selects that column
until this release is live.

Then run `cmd/backfill-talent-handle` once, also after the deploy. Migration `0148`
rewrites every retired `public` row to `anonymous` but cannot mint a handle — the
readable part comes from a Go dictionary — and `ListTalentNetworkMembers` requires
`talent_handle IS NOT NULL`. Without the backfill every pre-existing member is absent
from the catalogue and their card 404s until they toggle the control again. It is
idempotent.

## Review

Ran `/simplify` and `/code-review` (standards + spec axes). Everything they found is
fixed in the branch:

- **Existing members left with no address** — the backfill worker above. Found by the
  spec axis; it is the defect worth the whole exercise.
- **Migration number collision.** `0145` was free at the branch point, then
  `0145_mentorship` took it, then `0146`/`0147` went too. The chain moved twice and is
  `0148–0151` now. There is no gate and ordering falls to the alphabet.
- `Cache-Control` on a card was `public`, which authorises a shared cache to hold a
  departed member's card — this route's own promise broken by an intermediary. Now
  `private`.
- `min_years` was bounded `0..MaxInt32` in Go and `0..60` in TypeScript, so `min_years=100`
  was dropped by the UI while the API read it. Both are 60, and above it the parameter is
  reported rather than served an empty catalogue.
- The label maps are **exception tables** — `SENIORITY_LABELS` carries one entry — and
  reading them as complete dictionaries rendered every heading as a bare slug
  ("senior backend"). Surfaced by extracting a three-way duplicate and testing it.
- The mint loop moved out of the handler into `talentnetwork.Join`; `handler/AGENTS.md`
  says handlers are thin, and it still described the deleted route.

## Verification

`gofmt`, `go build`, `go vet`, `go vet -tags=integration`, `go test ./...`,
`pnpm --dir web check|lint|build`, 1717 vitest tests, `check:sql`, `check:links` and the
design-system adoption ratchet all pass. Merged with current `main`.

**Not done:** a manual browser click-through against a running app (task 7.3). Worth
doing before merge — that is the task the previous Talent Network change skipped, and how
the unreachable toggle shipped in the first place.

## Summary by CodeRabbit

* **New Features**
  * Added a public, anonymous Talent Network catalogue with filterable listings, pagination, and individual candidate cards.
  * Added stable handle-based profile links showing professional skills, roles, experience, locations, and specializations without personal names or free-text CV content.
  * Added Talent Network access to account navigation, profile invitations, and join/leave controls.
  * Added filtering by specialization, seniority, skills, location, timezone, and experience.

* **Changes**
  * Replaced the former public profile-link model with `off` and `anonymous` membership.
  * Added rate limiting and consistent not-found responses for unavailable cards.


strelov1:
**Deploy note updated: the DROP COLUMN is no longer in this PR.**

It carried a header saying to apply it after the deploy. `release.sh` does not read headers — it applies every pending migration and *then* flips the app, so the column would have been dropped out from under the server still selecting it. An instruction the deploy path never reads is not a safeguard.

So this PR is now ordinary: apply `0148`–`0150` the usual way, deploy, then run `cmd/backfill-talent-handle` once. Dropping `talent_network_public_id` ships as its own follow-up when the code that reads it is off production.

Every *read* of the column is already gone here — the query, the handler, the wire shape and the route.

PR #2646, opened 2026-09-08
Reuse the candidate's own cover letter in auto-apply form fields

strelov1:
## Summary
- When an ATS form asks for a cover letter as a free-text field (`cover_letter_text`), auto-apply now prefers the candidate's own already-drafted letter for that job (`internal/candidate/coverletter`) over the generic short answer `LLMDrafter` produces
- Falls back to the existing generic drafter unchanged when no letter has been drafted yet for that (candidate, job) pair, or when the read fails
- `LetterReader` is satisfied directly by `*coverletter.Store` — no adapter type needed, the same structural fit `AtomReader` already has over `*experience.Store`

Full OpenSpec change (proposal/specs/design/tasks) at `openspec/changes/autoapply-reuse-cover-letter/`.

## Test plan
- [x] New unit/scenario tests: field recognition (`isCoverLetterTextField`), letter lookup degradation (nil letter / read error), and `ResolveWithDrafting` scenarios (letter reused + drafter never called, no letter falls back to drafter, unrelated free-text field ignores the letter store, nil `LetterReader` leaves behavior unchanged)
- [x] `gofmt -l .` clean

## Summary by CodeRabbit

- **New Features**
  - Auto-apply now reuses an existing tailored cover letter for matching free-text cover-letter fields.
  - Cover-letter fields are recognized across supported ATS forms.
  - Missing, unreadable, or blank saved letters continue through standard drafting.
  - Select and radio fields are handled through their normal option-matching flow.

- **Documentation**
  - Added guidance describing cover-letter reuse and fallback behavior.

- **Tests**
  - Added coverage for reuse, fallback scenarios, field recognition, blank letters, and unrelated form fields.


strelov1:
Went through both findings, verified against the code:

**1. Blank stored body treated as a valid answer** — confirmed. \`matchOption\` accepts an empty string as a match for a free-text field with no options, so a stored \`coverletter.Stored\` row with an empty \`Body\` would have been submitted as a blank answer for a *required* field instead of falling back to the generic drafter. Fixed in \`coverLetterAnswer\`: a blank (whitespace-only) body now returns \`ok=false\`. Added \`TestCoverLetterAnswer_ABlankStoredBodyIsNotAnAnswer\` and an integration-level \`ResolveWithDrafting\` regression test.

**2. Select/radio fields reusing the stored letter** — confirmed. \`isCoverLetterTextField\` matches purely on ID/label, with no \`Kind\` check, so a \`select\` field whose label happens to match the cover-letter vocabulary (e.g. a "Cover Letter" yes/no dropdown) was routed to letter-reuse — the letter's prose would essentially never match an offered option label, so \`matchOption\` would park the field where the generic drafter (which can actually pick a real option) would have had a shot at answering it. Restricted reuse to \`Kind == "text"\`/\`"textarea"\` in \`answerFor\`. Added \`TestResolveWithDrafting_ACoverLetterMatchingSelectFieldIgnoresTheStoredLetter\`.

Updated the OpenSpec capability spec with two new scenarios covering both cases. Merged \`main\` (one conflict in \`cmd/auto-apply/main.go\` against the already-merged Recruitee-stored-schema change — both add their own \`NewClient\`/\`WithStoredFormReader\` wiring, resolved to keep both). Full \`gofmt\`/\`go vet\`/\`go vet -tags=integration\`/\`go test ./...\` and \`openspec validate --strict\` all clean.

PR #2650, opened 2026-09-08
Show the ghost-job badge on the tracker's own board

strelov1:
## Summary
- `GET /me/tracking` never showed the ghost-job badge on a tracked job's card, even though `/jobs` and search already do — `internal/application/userjob/AGENTS.md` named this gap explicitly
- `jobview.Card` gains a `Ghost` field, attached the same best-effort way the other two listings already do (reusing `ghostEvidenceFor`, `ListJobGhostStamps` — not duplicating them)
- The tracker's own listing query deliberately never reads a job's description (a measured, tested optimization — `TestMeasureBoardLoad` guards it), and reality classification normally needs it. Fixed by reading `RealityClass` from Meilisearch instead of recomputing it — exactly what search's own `attachGhost` already does with a live hit — via a new `search.In("id", [...])` filter (mirroring the swipe deck's existing `NotIn`) against the specific tracked ids
- `Card` deliberately gets **no** `Reality` field: neither `/jobs` nor search actually serves `Reality` on a card today, only the job detail page does — this follows that existing convention

Full OpenSpec change (proposal/specs/design/tasks) at `openspec/changes/tracker-ghost-badge/`.

## Test plan
- [x] New tests: converged ghost criteria produce a signal, a quiet job stays nil, an orphaned application is skipped without touching the id batch, a nil search dependency no-ops, a Meilisearch error degrades rather than failing the request
- [x] `TestMeasureBoardLoad` (integration) re-run: 0% description bytes, 0.43 MB payload — well under the 1 MB ceiling, confirming this change adds no description read
- [x] `gofmt -l .` clean

## Summary by CodeRabbit

- **New Features**
  - Tracked-job cards now include an optional ghost-job badge when ghost signals are detected.
  - Ghost status is derived from available tracking and search data without exposing job descriptions.
  - Missing data or lookup failures no longer prevent tracked jobs from loading.
  - Ghost badges remain available when sufficient tracking evidence exists, even if search data is unavailable.

- **Documentation**
  - Updated tracking documentation and specifications to describe ghost badges and their behavior.


strelov1:
Went through both findings and also fixed the failing CI (unrelated to the findings — I'd forgotten to run \`make gen-contracts\` after adding \`Card.Ghost\`, which is what the \`contracts\` job was actually failing on).

**1. Query-count wording (quick win)** — confirmed and fixed: \`attachGhostToTrackedCards\` does add one new Meilisearch \`Search\` call (\`search.In("id", ids)\`), separate from the reused \`ghostEvidenceFor\`/\`ListJobGhostStamps\` helpers. Updated the proposal's Impact section to say so.

**2. Lookup-failure contract ambiguity** — confirmed, and it was a real inaccuracy, not just unclear phrasing. Checked the actual code: a \`ListJobGhostStamps\` failure does abort the whole attach pass (every card's \`ghost\` omitted), but a \`Search\` (reality-class) failure only makes *that one* criterion unreachable — the design doc already documented this correctly in one place, but the spec's scenario said "every card's ghost field simply omitted" for *any* lookup failure, which overclaims. Split into two accurate scenarios. The existing test only covered "not enough non-reality evidence either way" (stays nil), so per your suggestion I added a genuine evidence-only-convergence regression test (\`ats_absent\` + \`user_reports\` converging with Search failing) — ran it against real Postgres via testcontainers to confirm the behavior actually holds, not just that the wording now matches an assumption.

Full \`gofmt\`/\`go vet\`/\`go vet -tags=integration\`/\`go test ./...\` (including the new integration test against real Postgres) and \`openspec validate --strict\` all clean.

PR #2653, opened 2026-09-08
Recognize Russian requirement headings, measured against real postings

strelov1:
## Summary
- `internal/job/reqextract/AGENTS.md` named this gap: the heading vocabulary was Latin-alphabet, mostly English (Hungarian was the one exception)
- Added four Russian entries — "Требования"/"Мы ждём от вас" (required), "Мы предлагаем"/"Условия" (closing) — each verified against 500 real `tbank.ru` postings in the local dev DB, not translated by hand
- "Обязанности" (duties, 682 real occurrences) is deliberately excluded: it describes what the job does, not what's required of the candidate — same as English's "what you'll do" is already excluded

**Important finding from measuring before trusting the fix:** all 683 real "Требования" headings sampled are followed by `<p>` paragraphs, never a `<ul>`/`<ol>` list — and `Derive` only extracts from a list. So this vocabulary addition extracts **zero** requirements from `tbank.ru` today; the real blocker there is a structural (list vs paragraph) mismatch that affects any language, not a language gap. Stated plainly in `AGENTS.md`, in the OpenSpec proposal, and as an honest negative test case in the suite itself.

Confirmed with the user before proceeding: ship the vocabulary addition as scoped. Closing the `<p>`-vs-list structural gap in `Derive`'s walk is a separate, larger, explicitly out-of-scope change (touches the shared walk every language and source goes through) — written up as a new `AGENTS.md` limitation rather than folded in here.

Full OpenSpec change (proposal/specs/design/tasks) at `openspec/changes/reqextract-russian-headings/`.

## Test plan
- [x] New `TestDerive` cases: the vocabulary firing correctly when a list follows the heading (positive case), and the real `tbank.ru` shape (heading + paragraphs) correctly yielding nothing (honest negative case)
- [x] `gofmt -l .` clean

## Summary by CodeRabbit

- **New Features**
  - Added support for recognizing Russian headings when extracting posting requirements.
  - Russian requirement sections can now provide extracted items when followed by supported lists.
  - Added recognition for headings indicating requirements, offers, and conditions.

- **Limitations**
  - Requirements written as individual paragraphs rather than list items are not extracted, regardless of language.

- **Documentation**
  - Documented supported Russian headings and the paragraph-versus-list extraction limitation.


strelov1:
Went through all 3 findings, verified each rather than applying blind:

**1. Missing "ё" transliteration variant** — confirmed empirically (not just from the library's changelog): ran \`unidecode.Unidecode\` against the pinned go-unidecode v0.2.0 directly. \`"Мы ждём от вас"\` (the grammatically correct spelling, with ё) normalizes to \`"my zhdiom ot vas"\`, not \`"my zhdem ot vas"\` — only the informal spelling without the dots (very common in Russian web text) matched the existing vocabulary entry. Added the missing variant plus a regression test for the accented form.

**2. Wrong occurrence-count mapping** — confirmed, fixed: the four phrases are 683/631/683/158 in listed order, not 683/683/158/631. Matches both the code's own comments and design.md, which already had it right — only the proposal's summary sentence had the counts shuffled.

**3. "No reindex needed" could read as "nothing to do for existing rows"** — agreed this was underspecified. Clarified explicitly: existing \`jobs.requirements_derived\` values are NOT automatically refreshed by this change — that needs either a fresh ingest write or a manual \`cmd/backfill-requirements\` pass (started from the beginning, since its own \`FROM_ID\` cursor resumes rather than restarts). Noted why this change doesn't run that backfill itself: the one large verified real source (tbank.ru) yields zero new extractions from this vocabulary regardless (the pre-existing \`<p>\`-vs-\`<ul>\` structural gap this proposal already documents), so there's nothing to backfill toward yet on the only data actually measured.

Full \`gofmt\`/\`go vet\`/\`go vet -tags=integration\`/\`go test ./...\` and \`openspec validate --strict\` all clean.

PR #2656, opened 2026-09-08
Fix teamtailor's listing cap silently truncating a real board

strelov1:
## Summary
- Part of the `fullBoardListing`-marker audit this session started — the first wave of ~19-21 large ATS platforms is marked; the rest was unaudited
- Audited the 8 adapters `AGENTS.md` already flags as "restate the walk by hand": all 8 fail the bar, most sharing the same defect (a later-page fetch error `break`s to a partial success instead of failing the whole `Fetch`)
- `teamtailor` specifically had a **live, confirmed active bug**, not just a marker-eligibility gap: `ttMaxPages=100` was silently truncating real boards
- Live-probed the two largest real `teamtailor` boards in prod: `migen.teamtailor.com` reaches its natural end well under page 100 (unaffected), but `tantor.teamtailor.com` (2000 open postings — a suspiciously round number) still returned a full page of fresh links at pages 100 **and** 101, with its real end only past page 120. Real catalogue ≈ 2400-2580 postings; the adapter was silently capping it at ~2000 — a confirmed ~20-25% under-collection on at least this one live board
- Fixed by raising the cap to 1000 (a wide multiple over the measured real maximum) and, mirroring `taleo.go`'s own established pattern, turning both a later-page failure and reaching the cap into hard `Fetch` failures instead of silent partial successes. `teamtailor` now earns the `fullBoardListing` marker on the strength of this fix

**Explicitly out of scope:** the other 7 audited adapters (`hh`, `neogov`, `edjoin`, `workstream`, `bayt`, `peopleforce`, `gusto`) are NOT fixed here — confirmed with the user to prioritize `teamtailor`'s live bug first. The rest stay a follow-up, noted honestly in `AGENTS.md`. Also out of scope: the ~163 remaining unaudited adapters — ranked by real prod volume, `apploi` (~1.47M open postings, larger than `workday`'s own ~1.05M) is the largest unmarked, unaudited one and hasn't been touched.

Full OpenSpec change (proposal/specs/design/tasks) at `openspec/changes/teamtailor-listing-cap-fix/` — includes a new `full-board-listing-marker` capability spec, since this contract previously lived only in `AGENTS.md`/code comments.

## Test plan
- [x] `go test ./...` — no failures across the module, all pre-existing `teamtailor` tests pass unchanged
- [x] New tests: a later-page failure now fails `Fetch`, an "endless" fake (mirroring `taleo_test.go`'s `taleoEndlessFake`) proves the raised cap is a hard failure with exactly the expected call count, and the standard `FullBoardListingProviders` registry assertion every marked provider carries
- [x] `gofmt -l .` clean

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Teamtailor job listing reliability by preventing incomplete crawls from being reported as successful.
  - Duplicate-only pages no longer incorrectly signal the end of a listing.
  - Listing failures on later pages now fail the crawl instead of returning partial results.
  - Increased the pagination limit to support larger job boards while still detecting incomplete listings.

- **Documentation**
  - Documented requirements for verifying complete board listings and updated the covered provider list.


strelov1:
Went through all 3 findings, verified against the code and live data:

**1-2. \`newLinks==0\` computed after cross-page dedup isn't a valid emptiness proof** — confirmed, same class of bug already fixed for 5 other adapters in the merged \`fullboardlisting-hand-rolled-batch\` PR. Fixed here too: \`jobURLs\`' stop condition now reads the raw per-page link count (before dedup), not the count of newly-added links. Live-verified against both currently-crawled teamtailor boards (\`migen\`, \`tantor\`) that neither actually repeats a page past its real end — both answer a genuine empty page — so this closes the gap without regressing the "board echoes its last page" tolerance the original code was defending against (that tolerance turned out to have no real evidence behind it for this adapter). Updated the existing repeat-page test (\`TestTeamtailorStopsWhenPageYieldsNoNewLinks\` → \`TestTeamtailorFetchFailsWhenAPageRepeatsAnEarlierLink\`, now asserts failure) and added a duplicate-then-new regression test.

**3. Safety-ceiling wording conflict in the capability spec** — agreed, tightened: the requirement now states explicitly that a page/offset ceiling is permitted only when reaching it hard-fails the crawl, not as an alternate silent-success path.

Also merged \`main\` (this branch had fallen behind the batch PR) and reconciled \`AGENTS.md\`'s now-six-adapter wave list. Full \`gofmt\`/\`go vet\`/\`go vet -tags=integration\`/\`go test ./...\` and \`openspec validate --strict\` all clean.

PR #2661, opened 2026-09-08
Fail hard, not silently, when 6 hand-rolled adapter listings truncate

strelov1:
## Summary

- `hh`, `neogov`, `edjoin`, `workstream`, `peopleforce`, `gusto` are 6 of the 8 adapters that restate the paginated-listing walk by hand instead of using the shared `crawlPagedLinks`/`crawlAllPagedLinks` helpers (`internal/ingest/sources/AGENTS.md`). All 8 predate the `fullBoardListing` marker and, until now, none earned it: a later listing page failing silently ended the walk with whatever was gathered, reported as an ordinary success — the same shape as #2337 (an unverified pagination cap on solidjobs silently closed 110 live postings).
- Each of the 6 now fails `Fetch` when a later page fails to fetch/decode, or when its walk exhausts its own page-cap safety constant without ever reaching a structurally-proven end (a genuinely empty page, or — for `neogov`/`edjoin`/`workstream` — the source's own declared total/page count). All 6 now implement `fullBoardListing` and are covered by `FullBoardListingProviders`.
- No page-cap constant is raised — no live measurement here shows any of these six boards actually being truncated by its existing cap today, unlike `teamtailor-listing-cap-fix` (a separate, in-flight change) which raises `ttMaxPages` on the strength of a confirmed live probe.
- `bayt` and `teamtailor` are deliberately excluded from this batch: `bayt` carries no source-declared total and is already fragile to throttling; `teamtailor` has its own change in flight.
- Full OpenSpec artifact set under `openspec/changes/fullboardlisting-hand-rolled-batch/` (proposal, a new `full-board-listing-marker` capability spec, design, tasks), validated with `openspec validate --strict`.

## Test plan

- [x] `gofmt -l .` clean
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (full suite, including new page-cap-exhaustion and marker-registration tests per adapter, and updated later-page-failure tests for `edjoin`/`gusto`)
- [x] `openspec validate fullboardlisting-hand-rolled-batch --strict`

## Summary by CodeRabbit

* **Bug Fixes**
  * Listing retrieval for six additional job sources now fails instead of returning incomplete results when later pages cannot be fetched or pagination cannot confirm completion.
  * Page-limit exhaustion is now reported as an error unless the listing’s natural end or declared total is verified.
  * These sources are now included in complete-board processing, improving accuracy when closing jobs no longer present in a listing.

* **Documentation**
  * Updated source documentation and specifications to describe the completeness requirements.


strelov1:
@coderabbitai review

strelov1:
Went through all 4 findings individually, verifying each against the actual code and (where it mattered) against live production data rather than applying the suggested fixes mechanically.

**1-2. Dedup-based `added==0` isn't a valid emptiness proof (edjoin/workstream/gusto + hh/neogov/peopleforce)** — confirmed and fixed for `neogov`, `edjoin`, `workstream`, `peopleforce`, `gusto`: the empty-page check now reads the RAW per-page item count (before cross-page dedup), not the count of newly-kept items. Added a regression test per adapter that serves a non-empty duplicate-only page followed by a genuinely new posting, proving the walk doesn't stop early.

For `hh` specifically, applying the same fix would have been actively harmful: `hh`'s own comment documents that hh.ru repeats/clamps its last page for an out-of-range request, and dedup is how that's detected. I live-verified this against a currently-configured board (professional_role 96) and found something more serious than the dedup question — hh.ru's own search UI caps a query's reachable depth at ~2000 results *regardless of a role's true count* (that role alone has 6,657 results in the adapter's 7-day window, and hh.ru's own paging state caps its `lastPage` at index 19, exactly `hhMaxPages`). That means `hh` can never structurally prove completeness for a busy board — so I pulled it out of the `fullBoardListing` wave entirely rather than switch its stop condition. It's reverted to its original soft cap-exhaustion behavior and documented in `hhru.go`'s own comments.

**3. peopleforce silently drops a posting on detail failure** — confirmed and fixed. `peopleforce` has no `HydratingSource` fallback (unlike the other 5), so every crawl re-fetches every listed posting's detail, and a transient failure was dropping it entirely — exactly the risk the marker exists to prevent. Reused the existing `unreadableDetail`/`detailUnreadable` pattern already established for 8 other non-hydrating adapters (careerplug.go is the closest structural match), rather than inventing something new. Added regression tests for both the "marked unreadable" and "404 still drops" cases.

**4. workstream comment inaccuracy** — fixed, wording now matches the implementation (`done` seeded from `stated`).

Also updated the OpenSpec artifacts (proposal/design/spec/tasks) and `AGENTS.md` to reflect the corrected scope (5 adapters instead of 6, `hh`'s audited exclusion with the live measurement recorded). Full `gofmt`/`go vet`/`go vet -tags=integration`/`go test ./...` all clean.

PR #2672, opened 2026-09-08
Bring bayt to the fullBoardListing bar

strelov1:
## Summary

- `bayt` was deliberately excluded from the earlier `fullboardlisting-hand-rolled-batch` PR (#2661) over a documented throttling risk. Investigated that risk before deciding: it's tied specifically to the concurrent detail fan-out (`baytDetailWorkers`, already narrowed for exactly that reason) — the listing walk itself is single-threaded/sequential and carries no evidence of the same risk. This sandbox can't isolate the listing walk's live throttling behavior (Bayt's Akamai edge 403s plain `curl` on TLS-fingerprint grounds regardless of pacing), so the reasoning is structural, documented as such rather than overclaimed as a live measurement.
- `Fetch`'s listing walk now proves completeness the same way the six already-fixed adapters do: the empty-page proof reads the raw count of job-detail links (post-`baytJobID` filter, pre-dedup) rather than every anchor on the page (which is never empty due to navigation chrome). A later page failing, or reaching `baytMaxPages` without a genuinely empty page, is now a hard `Fetch` failure.
- `detail` now returns the established `unreadableDetail` marker on a failed-but-not-404/410 fetch instead of dropping the posting — `bayt` has no `HydratingSource`, so every crawl re-fetches every listed posting's detail, and a throttled 403 (the documented risk) would otherwise silently vanish an already-known, still-live posting.
- `bayt` now implements `fullBoardListing` and is eligible for the board-scoped unseen-job close.
- No page-cap constant raised — no live measurement shows `baytMaxPages` (50) being reached by a real country listing.
- Full OpenSpec artifact set under `openspec/changes/bayt-fullboardlisting-fix/` (proposal, a `MODIFIED Requirements` delta against the `full-board-listing-marker` capability, design, tasks), validated with `openspec validate --strict`.

## Test plan

- [x] `gofmt -l .` clean
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (full suite, including new later-page-failure, duplicate-only-page, page-cap-exhaustion, unreadable-vs-gone-detail, and marker-registration tests)
- [x] `openspec validate bayt-fullboardlisting-fix --strict`

## Summary by CodeRabbit

- **Bug Fixes**
  - Bayt listings now fail clearly when later pages cannot be retrieved or pagination reaches its safety limit without confirming completion.
  - Duplicate-only pages no longer prematurely stop collection.
  - Unreadable or incomplete job details are preserved for reporting instead of being silently discarded.
  - Invalid job details continue to be excluded appropriately.

- **Improvements**
  - Bayt is now recognized as a complete-board listing source.

- **Documentation**
  - Updated ingest guidance and specifications for Bayt listing completeness.


strelov1:
Went through both findings, verified against the code:

**1. \`detail\` still dropped content-shape failures on a 200 response** — confirmed, and it was a real gap: I'd only wired \`unreadableDetail\` into the transport-failure branch, leaving "no ld+json JobPosting" and "no resolvable employer" as plain drops. Since \`bayt\` is now trusted for board-scoped close, a site-wide markup change breaking either of those would silently look like every posting on the board having been taken down — exactly the failure mode the marker exists to prevent. Fixed both branches to return \`unreadableDetail\`; updated the two existing tests that asserted the old drop behavior (\`TestBaytDropsPostingWithNoCompany\` → \`TestBaytUnreadableDetailForACompanyLessPosting\`, \`TestBaytDropsDetailWithNoJobPosting\` → \`TestBaytUnreadableDetailWithNoJobPosting\`).

**2. \`AGENTS.md\`'s cap wording** — agreed, tightened. The literal "no artificial page/offset cap" phrasing read as prohibiting a safety cap outright, when the actual (and already correctly-stated-elsewhere) rule is that a cap is fine as long as reaching it fails the crawl rather than silently succeeding partially.

Full \`gofmt\`/\`go vet\`/\`go vet -tags=integration\`/\`go test ./...\` and \`openspec validate --strict\` all clean.

PR #2740, opened 2026-09-09
Stop an upgrade or downgrade from opening a second Stripe subscription

strelov1:
## Summary

- `CheckoutURL` always opened a new Stripe Checkout Session, even for a customer who already had an active entitling subscription — so upgrading (or downgrading) tiers left two Stripe subscriptions billing the same customer in parallel instead of replacing one with the other. A real customer hit this in production (Pro $5/mo + Ultra $19/mo both active at once).
- `CheckoutURL` now reads the customer's current subscriptions and, when one already entitles them, changes it in place via Stripe's subscription item price replacement with proration — never a second Checkout Session. The decision (`decideCheckoutTarget`) is a pure, DB-free function so it's unit-testable without a database or network call; `CheckoutURL` itself is additionally covered by a real-Postgres/stubbed-Stripe integration test.
- Refuses to guess which item to change when a subscription already carries more than one (a shape Stripe's own Customer Portal can produce) — surfaced as an error rather than silently mutating the wrong item.
- The billing overview (`/billing/subscription`) now reports `multiple_subscriptions: true` when a customer already has more than one active entitling subscription, so an existing duplicate is visible on `/my/plan` instead of silently hidden behind the best-entitling one — with a warning banner pointing at "Manage or cancel".
- `CheckoutURL`'s response now reports the discount that was actually applied (never applied on an in-place update), rather than always echoing back what was offered.

Full rationale: `openspec/changes/fix-subscription-upgrade-duplicate/{proposal,design}.md`.

## Test plan

- [x] `gofmt -l .` clean
- [x] `go build ./...`, `go vet ./...`, `go vet -tags=integration ./...`
- [x] `go test ./...` (unit)
- [x] `go test -tags=integration ./internal/identity/billing/... ./internal/api/handler/...` (real Postgres via testcontainers)
- [x] `svelte-check` (0 errors) and `eslint` clean on the changed frontend files
- [ ] Manual Stripe test-mode verification (subscribe to Pro, click "Upgrade to Ultra", confirm via the Stripe dashboard it's the same subscription) — **not run**: no Stripe test-mode credentials available in the environment this PR was prepared in. Recommend running this once before/after merge.
- Existing duplicated subscriptions (like the one that surfaced this bug) are **not** cleaned up automatically — needs manual cancellation via the Stripe Customer Portal or dashboard for anyone already affected.

## Summary by CodeRabbit

* **New Features**
  * Upgrading or downgrading an existing subscription now updates it in place with proration instead of creating a duplicate subscription.
  * New subscriptions continue through hosted checkout.
  * Billing overviews identify customers with multiple active subscriptions.
  * The subscription panel displays a warning when duplicate subscriptions are detected.

* **Bug Fixes**
  * Discount responses report only discounts actually applied.
  * Ambiguous subscription configurations are rejected instead of creating another subscription.
  * Subscription lookup failures now stop checkout rather than opening an additional subscription.

### DIFF
--- a/internal/api/handler/billing.go
+++ b/internal/api/handler/billing.go
@@ -183,7 +183,7 @@ func (h *billingHandlers) Checkout(c *fiber.Ctx) error {
 
 	// The price comes from the pricing page's monthly/annual choice. It is validated
 	// against the configured list inside the service — never trusted as sent.
-	url, err := h.billing.CheckoutURL(ctx, userID, c.Query("price"), discount)
+	url, applied, err := h.billing.CheckoutURL(ctx, userID, c.Query("price"), discount)
 	if err != nil {
 		// Either checkout is unconfigured, or the provider refused. Neither is something a
 		// candidate can act on, and a 404 lets the surface omit the offer rather than render
@@ -193,10 +193,12 @@ func (h *billingHandlers) Checkout(c *fiber.Ctx) error {
 	}
 	return c.JSON(fiber.Map{"data": fiber.Map{
 		"url": url,
-		// Which offer was applied, so the page can say so rather than leaving the buyer to
-		// work it out from the amount on the provider's page. Empty when there was none.
-		"discount_percent": discount.PercentOff,
-		"discount_source":  discount.Label,
+		// Which offer was actually APPLIED — not merely offered — so the page can say so
+		// rather than leaving the buyer to work it out from the amount on the provider's
+		// page. Empty when there was none, including when a tier change modified an existing
+		// subscription in place rather than opening a checkout: that path never applies one.
+		"discount_percent": applied.PercentOff,
+		"discount_source":  applied.Label,
 	}})
 }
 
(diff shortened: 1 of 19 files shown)


strelov1:
Addressed in 673aa320:
- **Critical** (`service.go:337-340`, failed `GetStripeCustomerID` treated as unbound): fixed exactly as suggested — the read now fails the whole call instead of silently opening a checkout that could create a second customer/subscription. Added a regression test (`TestCheckoutURLFailsRatherThanTreatAFailedBindingReadAsUnbound`).
- `overview_test.go:195`: fixed — the fixture now actually sends both tiers' items on one subscription, matching what the test's own docstring claims to guard.
- `specs/subscription-billing/spec.md:8`: fixed — qualified the "exactly one entitling subscription afterward" postcondition to the case that had at most one beforehand, and added a requirement documenting the already-duplicated case's real (non-retroactively-fixed) behavior, matching `design.md`'s stated Non-Goal and the existing `TestDecideCheckoutTarget` "two concurrent entitling subscriptions" case.

**Not applied:** `client.go:276`, switching to `proration_behavior=always_invoice` + `payment_behavior=error_if_incomplete`. That's a product decision about *when* an upgrade charges the prorated difference (immediately vs. next cycle) — `design.md`'s Decisions section already deliberately chose the provider's default (`create_prorations`, deferred to the next invoice), and this PR's scope is stopping the duplicate-subscription bug, not changing upgrade billing timing. Worth a separate, explicitly-scoped change if immediate proration billing is wanted.

PR #2840, opened 2026-09-14
Report pool pressure without judging it, and restore the pagination wrapper

strelov1:
A quality pass over the deep-offset change (#2815, #2820, #2821). Two findings — the first is a defect I shipped.

## The defect: an instant cannot carry a day's verdict

`deriveSiteStatus` took pool pressure as an input and read `degraded` at 90% held, from a **single instant**.

But the pieces around it make that permanent. `StartSiteStatusSampler` takes **one reading every five minutes**, and `RecordSiteStatusSample` keeps the day's **worst** severity (`GREATEST`). Meanwhile the live pool, sampled every 5s against the healthy production site, reads:

```
9/10  10/10  0/10  0/10  3/10  0/10 ... 1/10 ... 2/10 ... 4/10  0/10
```

Real traffic is bursty and touching the ceiling is ordinary. One unlucky sample lands on a burst → the whole day is `degraded`, forever, and over 90 days the history strip goes yellow and stops meaning anything.

So pool pressure is **reported, never judged**. The number stays on the wire; the page shows it beside the error rate — unconditionally now, rather than as a warning above 90% that would have cried wolf on the same bursts. What turns it into a verdict is the Grafana rule, which averages it over five minutes (0.125–0.235 healthy against the outage's sustained 1.0). That needs history this process does not keep.

The parameter was also **shadowing** the `poolPressure` function, so neither could be called from inside the other.

## The tidying

`pageParamsWindowed` took the shared list caps at six of its seven call sites — exactly the repetition the `pageParams` wrapper existed to remove before this change orphaned it. Restored, now with the window folded in.

It returns an error where the old one returned two values, deliberately: a call site left behind by the change fails to compile rather than silently skipping the window.

## Verification

`gofmt` clean · `go vet ./...` clean · `go test ./...` 226 packages ok · `pnpm check` 0 errors · `pnpm test` 1943 tests.

Net **−39 lines**.

## Summary by CodeRabbit

- **Improvements**
  - Standardized pagination across company, feedback, jobs, search, and swipe-deck listings, with a shared deep-pagination limit.
  - Sitemap pages now use the same public-read rate limiting as other public endpoints.
  - Updated API documentation to clarify that excessive combined offset and limit values return an error.

- **Status Monitoring**
  - Site health status no longer treats connection-pool saturation as degraded.
  - Connection usage remains visible as a percentage in the status dashboard for monitoring.


strelov1:
## Code review закрыт

Two-axis review found 11 issues. Pushed a commit closing all of them; the two that mattered:

**Spec — a live hole of the same shape this change was fixing.** The requirement this change *added* reads:

> Every unauthenticated list endpoint SHALL mount a rate limiter. An endpoint registered without one is a defect regardless of how cheap its query is.

`sitemapHandlers.register` took **no middleware at all** and mounted four public routes bare. Same omission `/companies/:slug/feedback` carried — and invisible to the guard that caught that one, which only checks a register *mounting* a limiter is driven, so it structurally cannot see one mounting none. Now behind the public-read budget and driven by that guard.

Its offset stays unbounded, deliberately: a sitemap consumer must walk the whole index, and the pages come from Meilisearch, not a table scan. The two comments claiming "every list endpoint" now say **public** list.

**Standards — three documents asserting things the code does not do.**

| | |
|---|---|
| `web/src/lib/types.ts` | shipped *"A saturated pool reads `degraded`"* as the wire contract — the behaviour the simplify pass removed and argues against by name |
| `poolmetrics.go` Help | pointed operators at `freehire_db_pool_acquire_wait_seconds_total`, which the rename deleted |
| `openapi.yaml` | asserted `maximum: 10000` on offset — `offset=10000` with any limit is already a 400, and a wrong bound a generator trusts is worse than none |

**The rest:** the archived `tasks.md`/`proposal.md` claimed the pool signal became a `degraded` verdict (it did not — both now record the reversal, which is the more interesting half); four dangling `maxSearchWindow` references; `pageParamsWindowed`'s doc comment had merged into `pageParams`', leaving the actual guard undocumented; `observability/AGENTS.md` gained the six new metric families it was missing, including the three-draft alert table; `TestDurationMetricLabelsAreBounded` (the histogram was the one unbounded-label risk with no guard, and the one where a stray label costs ~7,700 series rather than ~700); `perf/k6/README.md` documents `deepoffset.js`.

**No scope creep found** — the proposal's "What does NOT change" list held.

`gofmt` clean · `go vet` clean · `go vet -tags=integration` clean · `go test ./...` 226 packages · `pnpm check` 0 errors · `pnpm test` 1943 tests.

PR #2876, opened 2026-09-15
Stop the heavy-index lock releasing itself forever

strelov1:
## The bug

`HoldHeavyIndexLock` returned a closure that called its own name:

```go
release, ok, err = holdHeavyIndexLock(ctx, l)   // named return value
...
return func() { release(); conn.Release() }, true, nil
```

`release` is the function's **named return value**. The `return` statement rebinds
it to the new closure *before anything calls it* — so the closure calls **itself**,
until the stack runs out.

## What it cost

Every run that **took** the lock died at the moment it let go — having already
finished all of its work:

| worker | what it does |
|---|---|
| `cmd/reindex` | rebuilds the Meilisearch jobs index |
| `cmd/reindex-companies` | rebuilds the companies index |
| `cmd/build-suggestions` | rebuilds the search box's dictionary, daily |

Shipped in #2864, live on production since `c1be0c461`.

## How it surfaced

CI on an **unrelated** pull request: `fatal error: stack overflow` in `pr-smoke`,
with `heavyindex.go:52` repeating down the whole trace. Nothing in the change under
test touched this code — `pr-smoke` passes on every other recent PR, which is what
made it worth reading rather than re-running.

## The fix

`releaseThen` takes the two functions as **ordinary arguments**, which a return
statement cannot reassign, and is testable without a database.

The order it fixes is load-bearing in its own right: the advisory lock is a
**session** lock living on that one connection, so unlocking *after* the connection
went back to the pool would be unlocking on whichever session takes it next.

## Verification

- New test asserts each function runs **exactly once** and unlock runs **first**
- **Mutation-verified**: reversing the order fails it with
  `release order = [conn unlock], want [unlock conn]`
- `go test ./...`, `go vet ./...`, gofmt, golangci-lint, govulncheck clean


strelov1:
Дубликат: #2874 уже в main и чинит ровно это — `composeRelease(inner, cleanup)` вместо замыкания, присваиваемого в именованный возврат, плюс `TestComposeRelease_CallsInnerThenCleanupExactlyOnce`, покрывающий те же два свойства (каждая функция ровно один раз, unlock перед возвратом соединения).

Найдено независимо: этот баг уронил `pr-smoke` на не связанном с ним PR (#2873) как `fatal error: stack overflow` с повторяющимся `heavyindex.go:52`. Раз в main лежит равноценная починка — добавить тут нечего.

